PR #22, opened 2021-10-05
Try adding a couple CSS reset fixes

### LINKED ISSUE #7: Front-end margins around the content do not match the site editor
There are a couple extra front-end margins that don't currently sync up with the site editor. Both of these have Gutenberg issues filed, but I'm noting them here so we can keep track of them: 

---

- [x] **Extra spacing around the `body` on the front end:** 

Screen Shot 2021-10-05 at 8 08 30 AM

Gutenberg issue: https://github.com/WordPress/gutenberg/issues/35350

---

- [x] **Extra `blockGap` spacing on top of the header:**

Screen Shot 2021-10-05 at 8 08 50 AM

Gutenberg issue: https://github.com/WordPress/gutenberg/issues/35353

kjellr:
@jffng not sure if we should include these temporarily or just wait for fixes on the Gutenberg end? They're pretty glaring issues. I was sure to reference Gutenberg issues directly for each one. 

Fixes #7 

Before|After
---|---
Screen Shot 2021-10-05 at 14 49 07|Screen Shot 2021-10-05 at 14 49 00

### DIFF
--- a/style.css
+++ b/style.css
@@ -15,6 +15,43 @@ Twenty Twenty-Two WordPress Theme, (C) 2021 WordPress.org
 Twenty Twenty-Two is distributed under the terms of the GNU GPL.
 */
 
+/*--------------------------------------------------------------
+>>> TABLE OF CONTENTS:
+----------------------------------------------------------------
+
+    0.  CSS Reset
+    1.  Fonts
+
+----------------------------------------------------------------------------- */
+
+/* -------------------------------------------------------------------------- */
+/*  0. CSS Reset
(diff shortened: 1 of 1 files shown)


jffng (commented):
I am torn, but I think we should work on the fixes upstream in Gutenberg first by suggesting a PR there. Even if the PR is naive, it's better than patching it here because another theme developer will eventually run into these problems.

kjellr:
This has since been resolved in Gutenberg, so I'm closing now. 👍

PR #36, opened 2021-10-06
Ensure only a string or false can be passed to main `wp_enqueue_theme`

### LINKED ISSUE #35: Fix phpstan issue with enqueue function
>phpstan is throwing the following error which I'm confused by, so I've had to `--no-verify` in order to commit:
>
> ```
> ------ -------------------------------------------------------- 
>  Line   functions.php                                           
> ------ -------------------------------------------------------- 
>  35     Parameter #4 $ver of function wp_enqueue_style expects  
>         bool|string|null, array|string|false given.             
> ------ -------------------------------------------------------- 
> ```

_Originally posted by @jffng in https://github.com/WordPress/twentytwentytwo/issues/26#issuecomment-937118508_

Because `wp_get_theme()->get()` can return `array|string|false` but the `wp_enqueue_style()` can only accept `bool|string|null` this error is flagged. We can either do a type check in advance to verify string or we can pass a static string.

I will open a PR with a typecheck and ternary that satisfies this error but the code doesn't look right. We might just want to flag it as ignored as the solution.

pattonwebz:
Fixes #35 by verifying in advance for static analysis tooling that only valid types (in this PR it constraints it to only a string or false) will be passed to the `$version` parameter of the `wp_enqueue_style()` function for the main stylesheet.

This doesn't seem like a perfect solution but it satisfies the static analysis tool.

An alternative option would be to simply have this code flagged for ignoring this error in the static analysis tool - which is the cleaner code approach and probably acceptable here since we know the `wp_get_theme->get( 'version' );` return value better than the docblock of the function details.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -19,13 +19,13 @@ if ( ! function_exists( 'twentytwentytwo_editor_styles' ) ) :
 	/**
 	 * Enqueue scripts and styles.
 	 */
-	function twentytwentytwo_scripts() {
+	function twentytwentytwo_styles() {
 		// Enqueue theme stylesheet.
 		$theme_version  = wp_get_theme()->get( 'Version' );
 		$version_string = is_string( $theme_version ) ? $theme_version : false;
 		wp_enqueue_style( 'twentytwentytwo-style', get_template_directory_uri() . '/style.css', array(), $version_string );
 	}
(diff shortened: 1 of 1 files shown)


scruffian:
FWIW I like this 👍

jffng (approved):
Thanks @pattonwebz

PR #38, opened 2021-10-06
Add gigantic and colossal font sizes to typography.fontSizes

### LINKED ISSUE #37: Add gigantic and colassal font sizes to typography.fontSizes
Is the a particular reason why the `gigantic` and `colossal` font sizes were omitted from `typography.fontSizes`? Or were they just looked over? 

They are added as ([custom css properties](https://github.com/WordPress/twentytwentytwo/blob/trunk/theme.json#L38-L39)), but not  accessible within the font size UI.

richtabor:
Closes https://github.com/WordPress/twentytwentytwo/issues/37 by pulling the previously custom-only gigantic and colossal font sizes, into the theme's fontSizes UI.

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -97,6 +97,16 @@
 					"name": "Huge",
 					"size": "clamp(2.5rem, 4vw, 3rem)",
 					"slug": "huge"
+				},
+				{
+					"name": "Gigantic",
+					"size": "clamp(3rem, 6vw, 4rem)",
+					"slug": "gigantic"
+				},
+				{
+					"name": "Colossal",
+					"size": "clamp(4rem, 8vw, 6.25rem)",
+					"slug": "colossal"
 				}
 			]
 		},
--- a/theme.json
+++ b/theme.json
@@ -34,10 +34,6 @@
 		},
 		"custom": {
 			"typography": {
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
The problem with this one is that if these presets are used in patterns, or on posts and the user changes theme, the font sizes disappear :D and the text is not gigantic or colossal anymore. If there's no preset for them, patterns can use the explicit value (though I'm not sure if they can use clamp right now in the editor) and that would survive a theme change

kjellr:
> The problem with this one is that if these presets are used in patterns, or on posts and the user changes theme, the font sizes disappear :D and the text is not gigantic or colossal anymore.

Yeah, that's generally the same thinking I noted here: https://github.com/WordPress/twentytwentytwo/issues/37#issuecomment-937784053

> If there's no preset for them, patterns can use the explicit value (though I'm not sure if they can use clamp right now in the editor) and that would survive a theme change

Yeah, patterns can use `clamp()` since Gutenberg considers that valid, but users can't add `clamp()` values without editing the HTML.

kjellr:
Closing this, since the associated issue has also been closed. Thanks for the PR!

PR #40, opened 2021-10-06
Replace h1 with paragraph in the footer site title

### LINKED ISSUE #39: Do not use H1 by default for footer site title
We should rather use a paragraph for this block by default.

audrasjb:
Fixes #39

Works fine on my side: the default h1 tag is replaced with a paragraph, which is more relevant in this context :)


kjellr (approved):
Thanks for catching that!

PR #43, opened 2021-10-06
Address floating issues on <main> element

### LINKED ISSUE #44: Address a floating issue on the <main> element
I believe we need to add an `overflow: hidden` declaration to the `<main>` element on each template to avoid floating issues. Not sure where is the best location for this CSS declaration though.

Issue:

screencapture-localhost-8888-twentytwentytwo-about-clearing-floats-2021-10-07-00_50_16

audrasjb:
Fixes #44 

I believe we need to add an `overflow: hidden` declaration to the `<main>` element on each template to avoid floating issues. Not sure where is the best location for this CSS declaration though.

Issue:

screencapture-localhost-8888-twentytwentytwo-about-clearing-floats-2021-10-07-00_50_16

After the proposed patch:

screencapture-localhost-8888-twentytwentytwo-about-clearing-floats-2021-10-07-00_48_51

### DIFF
--- a/block-templates/404.html
+++ b/block-templates/404.html
@@ -1,7 +1,7 @@
 <!-- wp:template-part {"slug":"header","tagName":"header","style":{"spacing":{"padding":{"right":"max(1.25rem, 5vw)","left":"max(1.25rem, 5vw)"}}},"layout":{"inherit":true}} /-->
 
 <!-- wp:group {"tagName":"main","style":{"spacing":{"padding":{"right":"max(1.25rem, 5vw)","left":"max(1.25rem, 5vw)"}}}} -->
-<main class="wp-block-group" style="padding-right:max(1.25rem, 5vw);padding-left:max(1.25rem, 5vw)"><!-- wp:group {"layout":{"inherit":true}} -->
+<main class="wp-block-group" style="padding-right:max(1.25rem, 5vw);padding-left:max(1.25rem, 5vw);overflow:hidden"><!-- wp:group {"layout":{"inherit":true}} -->
 <div class="wp-block-group"><!-- wp:heading {"style":{"typography":{"fontSize":"clamp(4rem, 40vw, 20rem)","fontWeight":"200","lineHeight":"0"}},"className":"has-text-align-center"} -->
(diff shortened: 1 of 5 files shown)


kjellr:
Thanks, @audrasjb. Is this an issue with Group blocks in general, or is there something specific to our templates that's causing the bug? If it's the latter, we should find a place for this change in Gutenberg.

audrasjb:
I would be a bit hesitant to implement this in the block itself as it depends on the context of use of the block…
In our case, we know the context (the main element), so at a glance, I'd say it's better to fix this in the theme :)

kjellr:
Right, but setting that block as the `main` element is a setting in Gutenberg — it could conditionally be applied there?

block-templates/404.html
```diff
@@ -1,7 +1,7 @@
 <!-- wp:template-part {"slug":"header","tagName":"header","style":{"spacing":{"padding":{"right":"max(1.25rem, 5vw)","left":"max(1.25rem, 5vw)"}}},"layout":{"inherit":true}} /-->
 
 <!-- wp:group {"tagName":"main","style":{"spacing":{"padding":{"right":"max(1.25rem, 5vw)","left":"max(1.25rem, 5vw)"}}}} -->
-<main class="wp-block-group" style="padding-right:max(1.25rem, 5vw);padding-left:max(1.25rem, 5vw)"><!-- wp:group {"layout":{"inherit":true}} -->
+<main class="wp-block-group" style="padding-right:max(1.25rem, 5vw);padding-left:max(1.25rem, 5vw);overflow:hidden"><!-- wp:group {"layout":{"inherit":true}} -->
```
youknowriad, line 4: Do we even support that style in Gutenberg? I mean won't this create an invalid block in the site editor?

kjellr: Yeah, it will. We'd need to include this elsewhere, but I do think this fix should be automatically supplied by GB.

kjellr:
Closing since this has been fixed upstream: https://github.com/WordPress/gutenberg/pull/35958

Thank you!

PR #45, opened 2021-10-07
Load editor styles correctly

### LINKED ISSUE #32: Source Serif Pro not being loaded in the editor
Currently the bundled font, Source Serif Pro, is not loaded in the site or post editor. No errors are thrown in the console.

nielslange:
Fixes #32 

## Note

At the moment, the editor styles are not loading. This is caused due to the function [`add_editor_style()`](https://developer.wordpress.org/reference/functions/add_editor_style/), which has a somewhat misleading name. This  function is not loading the styles for the block editor, but loads styles for the TinyMCE editor stylesheets.

I thought about two ways of resolving this problem:

- Switch hook `wp_enqueue_scripts` to `enqueue_block_assets`, to load the styles both in the frontend and editor.
- Switch hook `admin_init` to `enqueue_block_editor_assets` and replace `add_editor_style()` with `wp_enqueue_style()` within `twentytwentytwo_editor_styles()`.

I went for the first option, as it's less code, which improves both readability and maintainability of the source code.

<table>
<tr>
<td>Before:
<br><br>

#32-before
</td>
<td>After:
<br><br>

#32-after
</td>
</tr>
</table>

## Testing steps

1. Create a test page
2. Add the headlines 1-6 and save the page
3. See that Source Serif Pro is not rendered in the editor
4. Check out this PR
5. Refresh the page
6. See that Source Serif Pro is rendered in the editor

## Tip

The screenshot below shows where to see the rendered font within Chrome DevTools:

#32-rendered-font

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -11,6 +11,9 @@ if ( ! function_exists( 'twentytwentytwo_support' ) ) :
 
 		// Adding support for responsive embedded content.
 		add_theme_support( 'responsive-embeds' );
+
+		// Adding support for editor styles.
+		add_theme_support( 'editor-styles');
 	}
 	add_action( 'after_setup_theme', 'twentytwentytwo_support' );
 endif;
@@ -22,4 +25,17 @@ function twentytwentytwo_scripts() {
 	// Enqueue theme stylesheet.
 	wp_enqueue_style( 'twentytwentytwo-style', get_template_directory_uri() . '/style.css', array(), wp_get_theme()->get( 'Version' ) );
 }
-add_action( 'enqueue_block_assets', 'twentytwentytwo_scripts' );
+add_action( 'wp_enqueue_scripts', 'twentytwentytwo_scripts' );
+
+/**
+ * Enqueue editor styles.
+ */
+function twentytwentytwo_editor_styles() {
+	// Enqueue editor styles.
+	add_editor_style(
+		array(
+			get_stylesheet_uri(),
+		)
+	);
+}
+add_action( 'admin_init', 'twentytwentytwo_editor_styles' );
--- a/functions.php
+++ b/functions.php
@@ -22,17 +22,4 @@ function twentytwentytwo_scripts() {
 	// Enqueue theme stylesheet.
 	wp_enqueue_style( 'twentytwentytwo-style', get_template_directory_uri() . '/style.css', array(), wp_get_theme()->get( 'Version' ) );
 }
-add_action( 'wp_enqueue_scripts', 'twentytwentytwo_scripts' );
-
-/**
- * Enqueue editor styles.
- */
-function twentytwentytwo_editor_styles() {
-	// Enqueue editor styles.
-	add_editor_style(
-		array(
-			get_stylesheet_uri(),
-		)
-	);
-}
-add_action( 'admin_init', 'twentytwentytwo_editor_styles' );
+add_action( 'enqueue_block_assets', 'twentytwentytwo_scripts' );


functions.php
```diff
@@ -22,17 +22,4 @@ function twentytwentytwo_scripts() {
...
-	// Enqueue editor styles.
-	add_editor_style(
-		array(
-			get_stylesheet_uri(),
-		)
-	);
-}
-add_action( 'admin_init', 'twentytwentytwo_editor_styles' );
\ No newline at end of file
+add_action( 'enqueue_block_assets', 'twentytwentytwo_scripts' );
```
youknowriad: Won't this trigger a warning/error in the site editor. I believe all styles that need to be loaded in the editor need to be using editor styles. cc @ellatrix

nielslange: Using `enqueue_block_assets` instead of `wp_enqueue_scripts` doesn't trigger a warning/error in the site editor on my end. Based on [`enqueue_block_assets`](https://developer.wordpress.org/reference/hooks/enqueue_block_assets/), isn't this that hook that should be used in this particular situation? As mentioned before, the hook [`add_editor_style()`](https://developer.wordpress.org/reference/functions/add_editor_style/) affects TinyMCE editor stylesheets, but not the editor styles.

fabiankaegy: @nielslange I don't think that the `add_editor_style` is only ment for TinyMCE. It is specifically mentioned here in the Gutenberg docs: 

> ## Editor styles
>
>The block editor supports the theme's [editor styles](https://codex.wordpress.org/Editor_Style), however it works a little differently than in the classic editor.
>
>In the classic editor, the editor stylesheet is loaded directly into the iframe of the WYSIWYG editor, with no changes. The block editor, however, doesn't use iframes. To make sure your styles are applied only to the content of the editor, we automatically transform your editor styles by selectively rewriting or adjusting certain CSS selectors. This also allows the block editor to leverage your editor style in block variation previews.
>
>For example, if you write `body { ... }` in your editor style, this is rewritten to `.editor-styles-wrapper { ... }`. This also means that you should _not_ target any of the editor class names directly.
>
>Because it works a little differently, you need to opt-in to this by adding an extra snippet to your theme, in addition to the add_editor_style function:
>
>```php
>add_theme_support( 'editor-styles' );
>```
>
>You shouldn't need to change your editor styles too much; most themes can add the snippet above and get similar results in the classic editor and inside the block editor.
>
>### Enqueuing the editor style
>
>Use the `add_editor_style` function to enqueue and load CSS on the editor screen. For the classic editor, this was the only function needed to add style to the editor. For the new block editor, you first need to `add_theme_support( 'editor-styles');` mentioned above.
>
>```php
>add_editor_style( 'style-editor.css' );
>```
>
>Adding that to your `functions.php` file will add the stylesheet `style-editor.css` to the queue of stylesheets to be loaded in the editor.

https://github.com/WordPress/gutenberg/blob/9c3b78f290714a103573fdd9d294bfb0f9773cff/docs/how-to-guides/themes/theme-support.md#editor-styles 

There also is this explination from @ellatrix here: https://make.wordpress.org/core/2021/06/29/blocks-in-an-iframed-template-editor/#comment-41515

nielslange: Thanks for the additional information, @fabiankaegy. In this case, the explanation on https://developer.wordpress.org/reference/functions/add_editor_style/ is misleading.

Based on your feedback, I just updated this PR accordingly.

fabiankaegy (commented):
I tested this loacally and it appears to have fixed the issue in Chrome & Firefox. I am however still seeing the issue in Safari.

| Safari ⛔️ | Firefox ✅ | Chrome ✅ |
|-------|--------|---------|
|Screen Shot 2021-10-07 at 11 47 39 | Screen Shot 2021-10-07 at 11 47 47 | Screen Shot 2021-10-07 at 11 47 07|


## Browsers:
- [ ] Safari Version 15.0 (16612.1.29.41.4, 16612)
- [x] Firefox 93.0 (64-Bit)
- [x] Chrome Version 94.0.4606.71 (Official Build) (x86_64)

Because of this I am almost wondering wether there is an issue upstream in Gutenberg. I think the fix you made is correct and valid. But I think this will need a followup in Gutenberg

kjellr:
Thanks! @jffng and I originally took this out for a couple reasons:

- Removing it seemed to have no effect at all. Without it, I still see the same results @fabiankaegy [noted above](https://github.com/WordPress/twentytwentytwo/pull/45#pullrequestreview-773685890). 🤷 
- I noticed that in Core, `add_editor_style()` [appears to declare this itself anyway](https://core.trac.wordpress.org/browser/tags/5.8/src/wp-includes/theme.php#L2042). But now I'm seeing it says `add_theme_support( 'editor-style' )` without an `s` at the end. 🤔 So maybe that's something else?

I don't mind adding this back in, but I'm also not totally clear whether or not it's necessary.

jffng:
> I'm also not totally clear whether or not it's necessary.

Yeah, I don't understand the reason for adding this since #32 was resolved. I've tested with and without it and it doesn't appear to make a difference.

nielslange:
Closing this PR due to #50.

PR #46, opened 2021-10-07
Add SIL font license for Adobe’s Source Serif Pro

nickcernis:
This is required for redistribution. From condition 2 at https://github.com/adobe-fonts/source-serif/blob/main/LICENSE.md.

> Original or Modified Versions of the Font Software may be bundled, redistributed and/or sold with any software, provided that each copy contains the above copyright notice and this license. These can be included either as stand-alone text files, human-readable headers or in the appropriate machine-readable metadata fields within text or binary files as long as those fields can be easily viewed by the user.

Note, though, that both gnu.org and fsf.org list the SIL Open Font License 1.1 as incompatible with the GPL:

• https://www.gnu.org/licenses/license-list.html#SILOFL (orange left border means incompatible; see [legend](https://www.gnu.org/licenses/license-list.html#legend) at the top)
• https://directory.fsf.org/wiki/License:OFL-1.1
• https://www.gnu.org/licenses/gpl-faq.html#WhatIsCompatible

Presumably this is because, “1) Neither the Font Software nor any of its individual components, in
Original or Modified Versions, may be sold by itself”, which isn't a restriction the GPL seems to impose.

But the GNU analysis implies that requirement is “harmless” and satisfied by distributing with software (as we're doing here), and we're also not selling it.

> The Open Font License (including its original release, version 1.0) is a free copyleft license for fonts. Its only unusual requirement is that when selling the font, you must redistribute it bundled with some software, rather than alone. Since a simple Hello World program will satisfy the requirement, it is harmless. Neither we nor SIL recommend the use of this license for anything other than fonts.

### DIFF
--- a/assets/fonts/source-serif-pro/LICENSE.md
+++ b/assets/fonts/source-serif-pro/LICENSE.md
@@ -1,0 +1,93 @@
+Copyright 2014 - 2021 Adobe Systems Incorporated (http://www.adobe.com/), with Reserved Font Name 'Source'. All Rights Reserved. Source is a trademark of Adobe Systems Incorporated in the United States and/or other countries.
+
+This Font Software is licensed under the SIL Open Font License, Version 1.1.
+
+This license is copied below, and is also available with a FAQ at: http://scripts.sil.org/OFL
+
+
+-----------------------------------------------------------
+SIL OPEN FONT LICENSE Version 1.1 - 26 February 2007
+-----------------------------------------------------------
+
+PREAMBLE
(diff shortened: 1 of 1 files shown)


pattonwebz:
Historically themes have been able to bundle SIL licenced fonts in themes for the directory. Here is the list of acceptable font licences for reference: https://make.wordpress.org/themes/handbook/review/resources/#gpl-compatible-font-licenses

pattonwebz (approved):
LGTM

kjellr:
Thanks, folks!

PR #49, opened 2021-10-07
Opt-in to print styles for rendered blocks only and inlining them

aristath:
This PR adds `path` data to the main stylesheet, and enables support to only load styles for rendered blocks.
The result is that (depending on the blocks used in the content), the browser gets everything in a single request (or as few requests as possible), further improving performance.


aristath:
Rebased the PR to fix conflicts 👍

jffng (approved):
Merging this as we'll need it if / when any CSS is added.

PR #50, opened 2021-10-07
Remove .otf & .woff font-files

aristath:
Since WordPress no longer supports IE11, we don't need `.woff` and `.otf` font-files.
The `.woff2` format is superior, smaller in size, and supported by all browsers that WP supports (https://caniuse.com/woff2)

This PR removes the files, and their references in the CSS file.


jffng (approved):
Thank you!

PR #51, opened 2021-10-07
Fix font paths

aristath:
We can't assume that the theme will be in `wp-content/themes/twentytwentytwo`. Some sites use a different folder for wp-content, they have themes in subfolders, they might rename the theme and so on.
As a result, using URLs like `url('/wp-content/themes/twentytwentytwo/assets/fonts/source-serif-pro/SourceSerif4Variable-Roman.ttf.woff2')` in the stylesheet has a potential to break things.

A solution to that issue would be to add the font-face styles inline. This will allow us to use PHP to get the correct URL to the theme, and avoid breaking things.

Notes:
* This PR changes `twentytwentytwo-style` from `get_template_directory_uri() . '/style.css'` to an empty string (`''`). Since the stylesheet is completely empty and only contains the headers for the theme, loading it would be a waste of resources. Using an empty string allows us to inline styles using `wp_add_inline_style`, without actually enqueueing a file.
* This PR also removes non-`woff2` files. Since we no longer support IE11, these are not needed (also see #50)

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -20,8 +20,34 @@ if ( ! function_exists( 'twentytwentytwo_styles' ) ) :
 	 * Enqueue scripts and styles.
 	 */
 	function twentytwentytwo_styles() {
-		// The @font-face styles.
-		$font_face_css = "
+		// Register theme stylesheet.
+		wp_register_style( 'twentytwentytwo-style', '' );
+		// Add styles inline.
+		wp_add_inline_style( 'twentytwentytwo-style', twentytwentytwo_get_font_face_styles() );
+		// Enqueue theme stylesheet.
+		wp_enqueue_style( 'twentytwentytwo-style' );
+	}
+	add_action( 'wp_enqueue_scripts', 'twentytwentytwo_styles' );
+endif;
+
+if ( ! function_exists( 'twentytwentytwo_editor_styles' ) ) :
+	/**
+	 * Enqueue editor styles.
+	 */
+	function twentytwentytwo_editor_styles() {
+		wp_add_inline_style( 'wp-block-library', twentytwentytwo_get_font_face_styles() );
+	}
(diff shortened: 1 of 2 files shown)


jffng:
Thanks @aristath, curious what others think. 

Also I'm hoping we will be able to leverage https://core.trac.wordpress.org/ticket/46370#comment:53 so we can declare fonts via theme.json.

kjellr:
Interesting approach! I think this works for now, though we'll need to remember to re-enqueue the actual `style.css` file if we end up needing to pull in any CSS later on. (I'm keeping my fingers crossed that we won't!)

functions.php
```diff
@@ -19,8 +19,30 @@ function twentytwentytwo_support()  {
...
+		font-stretch: normal;
+		src: url('" . get_theme_file_uri( 'assets/fonts/source-serif-pro/SourceSerif4Variable-Roman.ttf.woff2' ) . "') format('woff2');
+	}
+
+	@font-face{
+		font-family: 'Source Serif Pro';
+		font-weight: 200 900;
+		font-style: italic;
+		font-stretch: normal;
+		src: url('" . get_theme_file_uri( 'assets/fonts/source-serif-pro/SourceSerif4Variable-Italic.ttf.woff2' ) . "') format('woff2');
```
westonruter: Why not just have the URL start with an `assets/` relative path? See https://github.com/WordPress/twentytwentytwo/pull/57

aristath: Small CSS files can be inlined (see #49 for example). If the `style.css` file gets inlined, relative paths are no longer relative to the `style.css` path but to the site's root - and in that case `assets/` won't work.

aristath:
Rebased the branch and there are no more conflicts 👍

jffng (approved):
This fixes font #32 also. Let's go with this for now, thanks @aristath.

PR #52, opened 2021-10-07
Remove superfluous function_exists checks

Ismail-elkorchi:
There is no need to make the functions pluggable, since child themes can modify or remove them using `remove_action/add_action`.

For further background please see : https://github.com/WordPress/twentyseventeen/pull/419, https://github.com/Automattic/_s/issues/1116, https://github.com/Automattic/_s/pull/1277

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -1,42 +1,36 @@
 <?php
 
-if ( ! function_exists( 'twentytwentytwo_support' ) ) :
-	function twentytwentytwo_support()  {
+function twentytwentytwo_support()  {
 
-		// Adding support for featured images.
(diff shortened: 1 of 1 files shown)


jffng:
Let's leave these in for now. Ideally we remove most of `functions.php` by the end of the theme's development anyhow. Thanks @Ismail-elkorchi

Ismail-elkorchi:
> Ideally we remove most of functions.php by the end of the theme's development anyhow.

Even though most of the functions.php file will be removed, this PR will still apply to the remaining part.  The future aspect of the code should not prevent us from sticking to good practices now.

PR #55, opened 2021-10-07
Use "header" template part inside of the "header-large-dark" template part

### LINKED ISSUE #54: Edited Navigation doesn't sync between the two header template parts
In order to provide the reversed-out header for our front page template, we need to use two separate header template parts: 

`header-large-dark.html`|`header.html`
---|---
Screen Shot 2021-10-07 at 11 59 37 AM|Screen Shot 2021-10-07 at 12 05 20 PM

This is great and it works fine. Except for one thing: 

If you edit the navigation block on one of the two headers, they don't sync up. In general, I think they should — if someone were to edit their main navigation, they presumably want it to change everywhere on their site. 

I _think_ the best way to fix this today is to create another "Navigation" template part containing just the Navigation block, and place that new template part inside both of our header template parts. It's a little confusing, but I believe it'll solve the problem. 

Open to other suggestions!

kjellr:
Fixes https://github.com/WordPress/twentytwentytwo/issues/54. 

This potential solution is thanks to a conversation I just had with @jasmussen. Rather than create a separate template part for the navigation block, we can actually just the the more generic `header.html` template part _within_ the `header-large-dark.html` template. 

When folks edit the navigation inside there, it'll update the original `header.html` template, and keep things in sync. 

The only issue I'm seeing is that the Site Title color doesn't appear to inherit the correct white link color (even though it's properly set on the container). I also tried setting it on the template part itself, but that doesn't seem to fix it either. Seems like a Gutenberg bug? 

Editor|Front end
---|---
Screen Shot 2021-10-07 at 12 53 46 PM|Screen Shot 2021-10-07 at 12 53 57 PM

## To test:

1. Using the site editor, edit the navigation while viewing the `front-page` template. 
2. Save your changes.
3. Ensure that change is synced to the navigation on other templates as well.

### DIFF
--- a/block-template-parts/header-large-dark.html
+++ b/block-template-parts/header-large-dark.html
@@ -1,15 +1,5 @@
-<!-- wp:group {"style":{"elements":{"link":{"color":{"text":"var:preset|color|background"}}},"spacing":{"padding":{"top":"max(1.25rem, 5vw)","right":"max(1.25rem, 5vw)","bottom":"max(1.25rem, 5vw)","left":"max(1.25rem, 5vw)"}}},"backgroundColor":"foreground","textColor":"background","layout":{"inherit":true}} -->
-<div class="wp-block-group has-background-color has-foreground-background-color has-text-color has-background has-link-color" style="padding-top:max(1.25rem, 5vw);padding-right:max(1.25rem, 5vw);padding-bottom:max(1.25rem, 5vw);padding-left:max(1.25rem, 5vw)"><!-- wp:group {"align":"wide","layout":{"type":"flex","justifyContent":"space-between"}} -->
-<div class="wp-block-group alignwide"><!-- wp:group {"layout":{"type":"flex"}} -->
-<div class="wp-block-group"><!-- wp:site-logo {"width":64} /-->
-
-<!-- wp:site-title /--></div>
-<!-- /wp:group -->
-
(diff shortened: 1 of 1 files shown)


jasmussen:
Nice one. I'll see if I can push a separate fix got the site title color inheritance to the block editor, so it probably shouldn't block this PR.

jffng (commented):
> Ensure that change is synced to the navigation on other templates as well

This aspect of the PR works as advertised. I did run into a visual bug though where the nav links are stacking strangely, but perhaps that's something to address separately: 

Screen Shot 2021-10-07 at 3 59 29 PM

kjellr:
That's weird! Are you seeing that without this PR too?

jffng:
> Are you seeing that without this PR too?

Yes, so maybe let's 🚢  and follow up.

jffng (approved):

jasmussen:
What a beautiful theme. And by the way, this header template will look so much better once [this PR](https://github.com/WordPress/gutenberg/pull/35397) lands:

Screenshot 2021-10-08 at 10 04 50

That said, I can't reproduce the issue. I'm correctly seeing the white color on the frontend and in the editor:
Screenshot 2021-10-08 at 10 06 32

I'm on latest gutenberg trunk. It's possible it's been fixed there inadvertently?

PR #56, opened 2021-10-07
Add a WordPress credit block pattern and load it in the footer

### LINKED ISSUE #30: String translation in templates
There are currently a few strings in the block templates that we will need to figure out what to do with from a translation / i18n perspective. Let's use this issue to discuss and track what to do there: 

- [x] Footer Credit
- [x] 404

Related Gutenberg issues / explorations:

- https://github.com/WordPress/gutenberg/issues/27402
- https://github.com/WordPress/gutenberg/pull/33217
- https://github.com/WordPress/gutenberg/pull/33192

jffng:
This PR is an attempt to solve #30. It depends on https://github.com/WordPress/gutenberg/pull/33217/

It adds a WordPress credit block pattern — which contains some text that can be translated — and includes it in the footer, instead of using the block markup directly.

**To test**
- Check out the above Gutenberg PR
- Verify the credit in the footer appears as expected in the site editor and front-end
- Verify you can add the pattern in the editor
- Verify you can edit it in the site editor

https://user-images.githubusercontent.com/5375500/136443900-684a0450-ec02-494a-ba70-d5d407d321ec.mp4

### DIFF
--- a/block-template-parts/footer.html
+++ b/block-template-parts/footer.html
@@ -1,7 +1,7 @@
 <!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"bottom":"max(1.25rem, 5vw)","top":"8rem","right":"max(1.25rem, 5vw)","left":"max(1.25rem, 5vw)"}}},"layout":{"type":"flex","justifyContent":"space-between"}} -->
 <div class="wp-block-group alignwide" style="padding-top:8rem;padding-right:max(1.25rem, 5vw);padding-bottom:max(1.25rem, 5vw);padding-left:max(1.25rem, 5vw)"><!-- wp:site-title {"level":0} /-->
 
-<!-- wp:paragraph {"align":"right"} -->
-<p class="has-text-align-right">Proudly Powered by <a href="https://wordpress.org" rel="nofollow">WordPress</a></p>
-<!-- /wp:paragraph --></div>
+<!-- wp:pattern {"slug":"twentytwentytwo/wordpress-credit"} /-->
+
+</div>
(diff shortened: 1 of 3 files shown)


kjellr:
<strike>Interesting idea! I'm seeing this issue across both of my test sites though:</strike>

Screen Shot 2021-10-07 at 2 49 33 PM

Edit: Whoops, I missed that this requires https://github.com/WordPress/gutenberg/pull/33217

kjellr:
Ok now that I've got it working, this is clever! I have one chief concern though: we shouldn't show these translation patterns in the inserter:

Screen Shot 2021-10-07 at 3 06 07 PM

I believe the two strings we have in the theme currently are for the footer, and for the 404 page. And it'll look _really_ confusing for folks to see "This page can not be found" text in the pattern inserter. They'll 100% think there's a bug. 

So if https://github.com/WordPress/gutenberg/pull/33217 lands, I think we'll also need some way to hide this sort of pattern. Given that, I worry a little bit about the "workarounds on top of workarounds" aspect of this approach. 😅

LittleBigThing:
Just wondering how this solution would work with multilingual websites? Probably it depends on the plugin that is used, but I assume most popular plugins do (can?) not support this.

jffng:
> Just wondering how this solution would work with multilingual websites? 

Since patterns are php files, the strings in those files can be marked for translation just like any other string supplied by a theme or plugin. The site may provide a .po/.pot file for each language, or use a plugin that provides a mechanism to translate + save translated strings.

jffng:
Closing in favor of #182.

PR #57, opened 2021-10-07
Use relative path instead of root-absolute path for font src

westonruter:
This fixes the path to the font files on setups where `/wp-content` is not the `WP_CONTENT_DIR`. For example, for me I have `/content` so that actual path would need to be `/content/themes/twentytwentytwo/assets/fonts/source-serif-pro/SourceSerif4Variable-Roman.ttf.woff2`. Making the path relative to the current directory fixes the issue.


westonruter:
Closed per https://github.com/WordPress/twentytwentytwo/pull/51#discussion_r724770826

Issue #61, opened 2021-10-07
Add CSS reset


colorful-tones:
I know we're trying to keep CSS mostly out of the theme, but likely a reset will be necessary. Unless we want to try to do something more upstream: Gutenberg or WordPress core integration?

I would recommend Andy Bell's (@andy-piccalilli) [`modern-css-reset`](https://github.com/andy-piccalilli/modern-css-reset). 

Thoughts? Feedback?

carolinan:
The upstream issue and discussions are here:
https://github.com/WordPress/gutenberg/issues/27116

https://github.com/WordPress/gutenberg/pull/27517#issuecomment-743181613
> Having a reset can break styles especially for things that are supposed to be "context agnostic". For example if you say `box-sizing: border-box` is part of the reset, blocks which styled are based on that assumption won't work on themes (or other contexts outside WP) that don't have that reset. For me, this kind of styles should be added to the blocks themselves and not assumed to be present. The theme shouldn't have to define the reset for these.

colorful-tones:
@carolinan thanks for linking up those upstream conversations. Very helpful.

I feel like this could likely be elevated and recommended as a WordPress core feature alongside an option in `theme.json` for themes to opt in/out. 

I’m willing to attempt a Trac ticket if there is ample momentum and foreseeable need. Thoughts?

carolinan:
It is not for me to decide -Can you show an issue that adding the reset would solve?

colorful-tones:
> It is not for me to decide

I know. 👍 

> Can you show an issue that adding the reset would would solve?

My initial problem was the `margin` on the `body`, but it seems likely there is already a fix for that in the works upstream: https://github.com/WordPress/gutenberg/pull/35421/files

However my next potential candidate for solving with a reset would be enabling a better experience for end users with respecting their choice for animations, e.g.

```
/* Remove all animations and transitions for people that prefer not to see them */
@media (prefers-reduced-motion: reduce) {
  html:focus-within {
   scroll-behavior: auto;
  }
  *,
  *::before,
  *::after {
    animation-duration: 0.01ms !important;
    animation-iteration-count: 1 !important;
    transition-duration: 0.01ms !important;
    scroll-behavior: auto !important;
  }
}
```

Yes, this presumes that animations are being used. Whether there is a significant number of themes that will likely use animation/transitions is relevant, and just trying to be forward thinking for end users _and_ theme developers.

kjellr:
[...]

PR #62, opened 2021-10-07
➕ ADD: modern-css-reset

colorful-tones:
Addresses #61

### DIFF
--- a/style.css
+++ b/style.css
@@ -20,6 +20,7 @@ Twenty Twenty-Two is distributed under the terms of the GNU GPL.
  ----------------------------------------------------------------
 
 	0. Fonts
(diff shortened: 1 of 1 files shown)


audrasjb:
+1 for this 👍

MaggieCabrera:
I'm not sure we need all of this on this particular theme. While a reset file is a good thing to have in general on a site, on a block theme there's a lot of this stuff that is already covered by the editor and the block themselves. I'd suggest we add these rules only when we need them, not by default just because.

kjellr:
Closing as per notes in https://github.com/WordPress/twentytwentytwo/issues/61#issuecomment-941591955.

PR #63, opened 2021-10-07
Add default feed links to <head>

audrasjb:
Just like any other bundled theme, Twenty Twenty-Two should insert posts and comments feed links to the `<head>` of the document by default :)

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -11,6 +11,9 @@ if ( ! function_exists( 'twentytwentytwo_support' ) ) :
 
 		// Adding support for responsive embedded content.
 		add_theme_support( 'responsive-embeds' );
+
+		// Adding default posts and comments RSS feed links to head.
+		add_theme_support( 'automatic-feed-links' );
 	}
 	add_action( 'after_setup_theme', 'twentytwentytwo_support' );
 endif;


kjellr:
This also seems like something that we should be able to cut out of the codebase for block themes. @youknowriad do you have any thoughts on this one?

audrasjb:
This is not the case for WordPress themes, so it would be kinda weird to have it by default for block themes and not for "classic" themes isn't it?

youknowriad:
I think we should take a look at the functions.php file of 2022 and potentially change the defaults for some of these theme supports for FSE themes (or themes with theme.json). What made sense as default for classic theme might not be true anymore in 2021. For instance, why does a theme still have to opt-in to HTML5 today. Why does it even need to be configurable.

I think we should take a look at all of these one by one and decide what makes most sense:

 - Make the theme support obsolete for FSE themes
 - Provide a theme.json config to replace it
 - opt-in or opt-out

Ideally, functions.php shouldn't be needed at all.

audrasjb:
I'm ok with this, but is it going to happen and realistic and properly documented in time for 5.9? If yes I'd definitely say "yes" to this change. But it needs to be properly announced :)

If we can't have a complete solution for 5.9, then I think we'd better stay with the current implementation of "classic" themes at least for Twenty Twenty-Two :)

jffng:
Since https://github.com/WordPress/gutenberg/pull/35593 merged, shall we close this one?

kjellr:
Yes, I think that makes sense. 👍

PR #64, opened 2021-10-07
Switch default core markup to output valid HTML5: styles & scripts declarations

audrasjb:
This PR fixes several HTML5 validation issues on script and styles HTML tags.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -11,6 +11,15 @@ if ( ! function_exists( 'twentytwentytwo_support' ) ) :
 
 		// Adding support for responsive embedded content.
 		add_theme_support( 'responsive-embeds' );
+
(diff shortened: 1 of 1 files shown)


kjellr:
Is this something that should just be added by default for all block themes?

audrasjb:
Not sure. This is currently an opt-in feature for theme developers so I'd say it would be weird to maintain two cases, one for themes and another for block themes.

kjellr:
For the record, a similar question came up a while back specifically regarding the comments block: https://github.com/WordPress/gutenberg/issues/30502.

jffng:
Since https://github.com/WordPress/gutenberg/pull/35593 merged, shall we close this one?

kjellr:
Yes, I think so. 👍

PR #65, opened 2021-10-08
add theme tags in style.css file

kafleg:


style.css
```diff
@@ -10,6 +11,7 @@ Version: 0.1
 License: GNU General Public License v2 or later
 License URI: http://www.gnu.org/licenses/gpl-2.0.html
 Text Domain: twentytwentytwo
+Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, featured-images, full-site-editing, block-patterns, rtl-language-support, sticky-post, threaded-comments
```
jffng, line 14: Is `sticky-post` still relevant?

kafleg: I am not sure but as this is a core feature, I think the theme should support it.

kjellr: Yeah, it's still a core feature, and it's supported by the Query block. We can include the tag here, but it seems like it's kind of obviously included given that this is a block-based theme. 

To be honest, most of these tags seem to describe to any block-based theme that's not aggressively locking down the block editor's built-in abilities. That being the case, I'm not sure they're all that useful for users. But that's more of a Theme Directory consideration I think. I think it's fine to add them in now and revise if the directory changes the way block themes are tagged later on.

kjellr (approved):
[As noted above](https://github.com/WordPress/twentytwentytwo/pull/65/files#r725640898), I think most of these tags are outdated for block themes, but since they're used in the Theme Directory, I think it makes sense to add them for now. We can revise later on if we'd like to. 

Thanks for the PR, @kafleg!

Issue #28, opened 2021-10-06
Internationalization hook not added into `after_setup_theme`


AbmSourav:
WordPress internationalization hook `load_theme_textdomain` is not initialized into `after_setup_theme` hook. 
Please let me know, I'll love to send a pull request :)

jffng:
👋 Depending on how #30 is handled, I am wondering whether we will need to use this function.

kjellr:
I think we probably do need it — if we do bundle any patterns with the theme, we'd need to use a textdomain. (Though that in turn depends on #12).

Otto42:
Calls to load the textdomain manually are not required and actually have not been since about WordPress 4.6 or so.

Short version: If you have the proper "Text Domain" and "Domain Path" headers in the style.css file, then WordPress will automatically load the text domain on a just-in-time basis, when they are needed.

The Text Domain should always be identical to the theme's slug, in this case, "twentytwentytwo".

The Domain Path should be the directory inside the theme's directory where any local language files may be located. The default is "/languages".

So, the given pull request that adds this line of code:

`load_theme_textdomain( 'twentytwentytwo', get_template_directory() . '/languages' );`

In fact that code does absolutely nothing. It's functionally the same as just having the Text Domain: header, which the theme already has.

kjellr:
Thanks for the background, @Otto42 — that's really helpful. I think we can close in that case.

Otto42:
Just adding this to note that this only applies to themes and plugins hosted on wordpress.org. Those get translated by the translation pack system. Third party code hosted elsewhere should include their own translations, as per usual.

PR #67, opened 2021-10-08
Load theme text domain

kafleg:


kjellr:
Hey @kafleg. As per the discussion in https://github.com/WordPress/twentytwentytwo/issues/28, this should not be necessary for the theme. I'm going to close for now, but we can reopen the issue if more discussion is needed. Thanks!

PR #69, opened 2021-10-08
Move front-page template to page-home instead.

### LINKED ISSUE #60: Homepage content & user experience
Perhaps I'm not familiar enough with Block Theme/FSE, but as an everyday editor I would be confused by the following dilemma. I created a new empty WordPress site in Local. I have TwentyTwentyTwo theme activated and latest Gutenberg.

I already have **44 posts** I've generated using [FakerPress plugin](https://wordpress.org/plugins/fakerpress/). I visit my site's homepage and see: "Welcome to WordPress. This is your first post. Edit or delete it, then start writing!" Here, the confusion begins.

My first instinct is to check my **Settings** > **Reading** and verify that "Your homepage displays: Your latest posts". This is contradicting what I'm seeing on the homepage though. 😕 

It is likely this is a user experience improvement for Gutenberg Full Site Editing experience, and just wondering if I'm missing something?

Screen Shot 2021-10-07 at 4 42 39 PM

Screen Shot 2021-10-07 at 4 41 19 PM

kjellr:
Fixes https://github.com/WordPress/twentytwentytwo/issues/60. This PR renames `front-page.php` to `page-home.php` so that the theme will respect the user's front page settings.


block-templates/home.html
```diff
@@ -0,0 +1,55 @@
...
+<main class="wp-block-group" style="padding-right:max(1.25rem, 5vw);padding-left:max(1.25rem, 5vw)">
+
+<!-- wp:group {"align":"wide","layout":{"inherit":false}} -->
+<div class="wp-block-group alignwide"><!-- wp:group {"layout":{"type":"flex","justifyContent":"space-between"}} -->
+<div class="wp-block-group"><!-- wp:paragraph -->
+<p>LATEST POSTS</p>
+<!-- /wp:paragraph -->
+
+<!-- wp:paragraph -->
+<p><a href="https://2022kjell.mystagingwebsite.com/blog/" data-type="page" data-id="8">View All</a></p>
```
Sandstromer: Not sure if leaving this hardcoded link is intentional.

kjellr:
I reworked this a bit (and updated the description). Originally I tried having a `home.html` template that had the large black header, but that [forced a lot of extra scrolling](https://cloudup.com/cTpeH0BHWdt) when you paginated. I might add a `home.html` template later on, but I'll need to redesign it so this is smoother.

This new approach forces folks to either name their homepage with the `home` slug, or to manually select the `page-home` template for their homepage. But I think that's more or less what the template hierarchy dictates. Maybe we can use starter content to make that easier.

block-templates/index.html
```diff
@@ -1,7 +1,7 @@
 <!-- wp:template-part {"slug":"header","tagName":"header","style":{"spacing":{"padding":{"right":"max(1.25rem, 5vw)","left":"max(1.25rem, 5vw)"}}},"layout":{"inherit":true}} /-->
 
 <!-- wp:group {"style":{"spacing":{"padding":{"right":"max(1.25rem, 5vw)","left":"max(1.25rem, 5vw)"}}}} -->
-<div class="wp-block-group" style="padding-right:max(1.25rem, 5vw);padding-left:max(1.25rem, 5vw)"><!-- wp:query {"queryId":1,"query":{"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":"","perPage":10},"tagName":"main","displayLayout":{"type":"flex","columns":3},"layout":{"inherit":true}} -->
+<div class="wp-block-group" style="padding-right:max(1.25rem, 5vw);padding-left:max(1.25rem, 5vw)"><!-- wp:query {"queryId":1,"query":{"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":"","perPage":12},"tagName":"main","displayLayout":{"type":"flex","columns":3},"layout":{"inherit":true}} -->
```
jffng, line 4: This may be a separate issue, but is there a way to get the query to respect the Reading settings for max blog posts to show?

Screen Shot 2021-10-08 at 11 49 09 AM

I think it's 10 by default.

kjellr: Yeah, I _think_ if you leave that out it will respect this setting. But we definitely want to have the default be a multiple of three though since we're showing things in three columns. It seems like setting that here is the best way to do this for a block theme. 🤷‍♂️

jffng (approved):

Issue #14, opened 2021-10-05
Block Patterns: Add Header Patterns


kjellr:
![Header patterns](https://user-images.githubusercontent.com/1202812/136079592-8e550fd9-740b-41a8-b118-d93a5ec11b9f.png)

These mockups are available with more detail in [the theme's Figma file](https://www.figma.com/file/76mfUcaK4QDlrXElk8MK3H/?node-id=758%3A4770).

kjellr:
Closed via #74.

Issue #12, opened 2021-10-05
Determine strategy for block patterns


kjellr:
In previous themes, block patterns were bundled directly with the theme itself. We should consider doing that here, but also keep an eye on Pattern Directory progress. Can we instead put these patterns in the directory?

In addition to opening these patterns up to users beyond this single theme, having our patterns in the directory would cut down our theme code considerably, ensure translation is handled automatically, and also make them easier to update on the fly. 

Related Gutenberg issue: https://github.com/WordPress/gutenberg/issues/35364

jffng:
It appears that this will not be in 5.9, so should we close this one since we will have to bundle them with the theme? 

In any case, we can upload them to the patterns directory, even if we can't remove them from the theme.

kjellr:
Yep, I think it's fine to close for now. 👍

PR #74, opened 2021-10-08
Add header block patterns

kjellr:
**Description**

Addresses #14. (_Probably_ closes it for now at least). As per #12, we may move these to the directory at some point, but I think it makes sense to start building them in here to get started. 

This PR adds 12 of the block patterns from the initial designs:

Header-Patterns

The four that are marked with an ❌ were not possible to create with the current block controls. I think that's fine though... we can leave them out for now. 

Some technical notes: 

- I added these to a "Twenty Twenty-Two Headers" category. We'll have a lot of block patterns in this theme, so I think it makes sense to have multiple, specific categories. 
- This includes the default `header` and `header-large-dark` patterns as used in our block template parts. Theme template parts don't currently show up in the theme carousel, so I think this is a sound strategy? That should probably change though.
- A few of these don't wrap super-nicely on smaller screens. 😕
    -  Header with Tagline ([Screenshot](https://cloudup.com/cxBvFm2Oj3V))
    - Text-only Header with Tagline ([Screenshot](https://cloudup.com/cz_z0CveuCt))
    - Text-only Header with Tagline and all-caps Title ([Screenshot](https://cloudup.com/c4mv-0BEeLj))
- <strike>The vertical gap between the title and tagline on [**Header with Tagline**](https://cloudup.com/c7ph2HPsD_O) and [**Text-only Header with stacked Tagline**](https://cloudup.com/c6CvKVhG5B5) is clearly too big. I'm not sure if we can adjust that yet though?</strike>
- I opted not to include any left/right padding around the patterns themselves. In our templates I've been adding that to the [parent template part](https://github.com/WordPress/twentytwentytwo/blob/af5d31d52c364670a09c8f2a85f77c9a592ca2bd/block-templates/single.html#L1), since it needs to match up with [the padding around the `main` template](https://github.com/WordPress/twentytwentytwo/blob/af5d31d52c364670a09c8f2a85f77c9a592ca2bd/block-templates/single.html#L3) too. Keeping left/right padding out of the patterns seems more versatile than having it in there. 
- While editing this I realized that the Site Title in our `header.html` and `header-large-black.html` template parts is supposed to be italic in order to match the comps. I made that change here to sync things up. 

**Screenshots**

block-patterns

**Testing Instructions** 

1. Add a "Header" block.
2. Choose "New header"
3. Scroll through the carousel, insert the new patterns, and test them out.

### DIFF
--- a/inc/patterns/header-centered-logo-in-navigation.php
+++ b/inc/patterns/header-centered-logo-in-navigation.php
@@ -6,13 +6,17 @@ return array(
 	'title'      => __( 'Header with centered Logo in Navigation', 'twentytwentytwo' ),
 	'categories' => array( 'twentytwentytwo-headers' ),
 	'blockTypes' => array( 'core/template-part/header' ),
-	'content'    => '<!-- wp:group {"align":"wide","tagName":"header","style":{"spacing":{"padding":{"bottom":"8rem","top":"max(1.25rem, 5vw)"}}},"layout":{"type":"flex","justifyContent":"center"}} -->
+	'content'    => '<!-- wp:group {"tagName":"header","align":"wide","style":{"spacing":{"padding":{"bottom":"8rem","top":"max(1.25rem, 5vw)"}}},"layout":{"type":"flex","justifyContent":"center"}} -->
 					<header class="wp-block-group alignwide" style="padding-top:max(1.25rem, 5vw);padding-bottom:8rem"><!-- wp:navigation {"itemsJustification":"right","isResponsive":true} -->
-					<!-- wp:page-list {"isNavigationChild":true,"showSubmenuIcon":true,"openSubmenusOnClick":false} /-->
+					<!-- wp:navigation-link {"isTopLevelLink":true} /-->
+
+					<!-- wp:navigation-link {"isTopLevelLink":true} /-->
 
 					<!-- wp:site-logo {"width":90} /-->
 
-					<!-- wp:page-list {"isNavigationChild":true,"showSubmenuIcon":true,"openSubmenusOnClick":false} /-->
+					<!-- wp:navigation-link {"isTopLevelLink":true} /-->
+
+					<!-- wp:navigation-link {"isTopLevelLink":true} /-->
 					<!-- /wp:navigation --></header>
 					<!-- /wp:group -->',
 );
--- a/inc/patterns/header-centered-logo-in-navigation.php
+++ b/inc/patterns/header-centered-logo-in-navigation.php
@@ -6,13 +6,13 @@ return array(
 	'title'      => __( 'Header with centered Logo in Navigation', 'twentytwentytwo' ),
 	'categories' => array( 'twentytwentytwo-headers' ),
 	'blockTypes' => array( 'core/template-part/header' ),
-	'content'    => '<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"bottom":"8rem","top":"max(1.25rem, 5vw)"}}},"layout":{"type":"flex","justifyContent":"center"}} -->
-					<div class="wp-block-group alignwide" style="padding-top:max(1.25rem, 5vw);padding-bottom:8rem"><!-- wp:navigation {"itemsJustification":"right","isResponsive":true} -->
+	'content'    => '<!-- wp:group {"align":"wide","tagName":"header","style":{"spacing":{"padding":{"bottom":"8rem","top":"max(1.25rem, 5vw)"}}},"layout":{"type":"flex","justifyContent":"center"}} -->
+					<header class="wp-block-group alignwide" style="padding-top:max(1.25rem, 5vw);padding-bottom:8rem"><!-- wp:navigation {"itemsJustification":"right","isResponsive":true} -->
 					<!-- wp:page-list {"isNavigationChild":true,"showSubmenuIcon":true,"openSubmenusOnClick":false} /-->
 
 					<!-- wp:site-logo {"width":90} /-->
 
 					<!-- wp:page-list {"isNavigationChild":true,"showSubmenuIcon":true,"openSubmenusOnClick":false} /-->
-					<!-- /wp:navigation --></div>
+					<!-- /wp:navigation --></header>
 					<!-- /wp:group -->',
 );
--- a/inc/patterns/header-centered-logo-in-navigation.php
+++ b/inc/patterns/header-centered-logo-in-navigation.php
@@ -6,8 +6,8 @@ return array(
 	'title'      => __( 'Header with centered Logo in Navigation', 'twentytwentytwo' ),
 	'categories' => array( 'twentytwentytwo-headers' ),
 	'blockTypes' => array( 'core/template-part/header' ),
-	'content'    => '<!-- wp:group {"tagName":"header","align":"wide","style":{"spacing":{"padding":{"bottom":"8rem","top":"max(1.25rem, 5vw)"}}},"layout":{"type":"flex","justifyContent":"center"}} -->
-					<header class="wp-block-group alignwide" style="padding-top:max(1.25rem, 5vw);padding-bottom:8rem"><!-- wp:navigation {"itemsJustification":"right","isResponsive":true} -->
+	'content'    => '<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"bottom":"8rem","top":"max(1.25rem, 5vw)"}}},"layout":{"type":"flex","justifyContent":"center"}} -->
+					<div class="wp-block-group alignwide" style="padding-top:max(1.25rem, 5vw);padding-bottom:8rem"><!-- wp:navigation {"itemsJustification":"right","isResponsive":true} -->
 					<!-- wp:navigation-link {"isTopLevelLink":true} /-->
 
 					<!-- wp:navigation-link {"isTopLevelLink":true} /-->
@@ -17,6 +17,6 @@ return array(
 					<!-- wp:navigation-link {"isTopLevelLink":true} /-->
 
 					<!-- wp:navigation-link {"isTopLevelLink":true} /-->
-					<!-- /wp:navigation --></header>
+					<!-- /wp:navigation --></div>
 					<!-- /wp:group -->',
 );
--- a/inc/patterns/header-centered-logo-in-navigation.php
+++ b/inc/patterns/header-centered-logo-in-navigation.php
@@ -3,10 +3,10 @@
  * Header with centered Logo in Navigation block pattern
  */
 return array(
-	'title'       => __( 'Header with centered Logo in Navigation', 'twentytwentytwo' ),
-	'categories'  => array( 'twentytwentytwo-headers' ),
+	'title'      => __( 'Header with centered Logo in Navigation', 'twentytwentytwo' ),
+	'categories' => array( 'twentytwentytwo-headers' ),
 	'blockTypes' => array( 'core/template-part/header' ),
-	'content'     => '<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"bottom":"8rem","top":"max(1.25rem, 5vw)"}}},"layout":{"type":"flex","justifyContent":"center"}} -->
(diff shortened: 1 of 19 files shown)


inc/patterns/header-logo-navigation-social.php
```diff
@@ -0,0 +1,22 @@
...
+	'title'       => __( 'Logo, Navigation, and Social Links Header', 'twentytwentytwo' ),
+	'categories'  => array( 'twentytwentytwo-headers' ),
+	'blockTypes' => array( 'core/template-part/header' ),
+	'content'     => '<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"bottom":"8rem","top":"max(1.25rem, 5vw)"}}},"layout":{"type":"flex","justifyContent":"space-between"}} -->
+					<div class="wp-block-group alignwide" style="padding-top:max(1.25rem, 5vw);padding-bottom:8rem"><!-- wp:site-logo {"width":64} /-->
+
+					<!-- wp:navigation {"itemsJustification":"right","isResponsive":true} -->
+					<!-- wp:page-list {"isNavigationChild":true,"showSubmenuIcon":true,"openSubmenusOnClick":false} /-->
+
+					<!-- wp:social-links {"iconColor":"foreground","iconColorValue":"#000000","className":"is-style-logos-only"} -->
```
jffng: Does it make sense to use a preset here? e.g. 
```suggestion
					<!-- wp:social-links {"iconColor":"foreground","iconColorValue":"var(--wp--preset--color--foreground)","className":"is-style-logos-only"} -->
```

inc/patterns/header-default.php
```diff
@@ -0,0 +1,21 @@
...
+	'blockTypes' => array( 'core/template-part/header' ),
+	'content'     => '<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"bottom":"8rem","top":"max(1.25rem, 5vw)"}}},"layout":{"type":"flex","justifyContent":"space-between"}} -->
+					<div class="wp-block-group alignwide" style="padding-top:max(1.25rem, 5vw);padding-bottom:8rem"><!-- wp:group {"layout":{"type":"flex"}} -->
+					<div class="wp-block-group">
+					<!-- wp:site-logo {"width":64} /-->
+
+					<!-- wp:site-title {"style":{"typography":{"fontStyle":"italic","fontWeight":"400"}}} /--></div>
+					<!-- /wp:group -->
+
+					<!-- wp:navigation {"itemsJustification":"right","isResponsive":true} -->
```
jffng, line 17: If I update the navigation in the standard header, and I go to insert a new header from one of these patterns, it pulls in the old / original navigation. Is there a way to get the pattern to pull in a navigation template part, i.e. the nav that I've updated already?

kjellr: Not yet! But there's some early work in https://github.com/WordPress/gutenberg/pull/35418 to help make that sort of thing happen. I suggest we hold on and see how that shakes out before implementing any sort of workarounds here.

inc/patterns/header-logo-navigation-social.php
kjellr: I think you've spotted a Gutenberg bug actually. 😄 

When you set a social link color using preset color, it embeds the hex value there for `iconColorValue`. And it continues using that hex value, even if you've changed the original preset color via Global Styles.

I pushed 65da3c1 to use our preset variables for now (this appears to work for our purposes), but I'll also file an upstream bug report.

kjellr: Filed here: https://github.com/WordPress/gutenberg/issues/35480

jffng: > it continues using that hex value, even if you've changed the original preset color via Global Styles.

Nice detective work.

kjellr:
> The vertical gap between the title and tagline on Header with Tagline and Text-only Header with stacked Tagline is clearly too big. I'm not sure if we can adjust that yet though?

Fixed in b047708 via some block margin adjustments. 👍 

Screen Shot 2021-10-11 at 8 14 20 AM

kjellr:
After chatting on slack with @mtias and @jasmussen, I've pushed mostly-working versions of the four patterns that were marked with an ❌ to this PR as well. I've added them to the end of the patterns list. Let's merge them in for now and open a followup issue to address issues. For now, here they are along with the known bugs: 

---

**Header with centered Logo**

Screen Shot 2021-10-11 at 9 54 18 AM

Current issues: 

- This wraps poorly on small screens. [(GIF)](https://cloudup.com/ciCs3nzlqFN)

---

**Header with centered Logo in Navigation**

Screen Shot 2021-10-11 at 9 57 01 AM

Current issues: 

- The site logo disappears on smaller screens. ([Screenshot](https://cloudup.com/c-krv3vU4u7))
- The layout of the expanded menu looks weird. ([Screenshot](https://cloudup.com/cBDKpdI0Fbz))
- Depending on the number/length of menu items, the logo is not necessarily located in the exact center. ([Screenshot](https://cloudup.com/cZzkpra6jFl))

---

**Centered title with Navigation and Social Links Header**

Screen Shot 2021-10-11 at 10 03 06 AM

Current issues: 

- The items wrap oddly on small screens. ([Screenshot](https://cloudup.com/c91b-Oh1-oD))
- The navigation items start stacking in a weird way on medium breakpoints. ([Screenshot](https://cloudup.com/cq9BAkOnkXk))

---

**Title and Button Header**

Screen Shot 2021-10-11 at 10 04 47 AM

Current issues: 

- This one is supposed to have the hamburger icon visible by default on all breakpoints. Requires a resolution to https://github.com/WordPress/gutenberg/issues/34783. 

---

This should be ready for a review. If they generally look good, I suggest we follow up with issues for everything above and work from there.

mtias:
The ones that stack awkwardly in mobile are good ways to test solutions for auto-collapsing containers.

scruffian:
Quite a lot of the patterns use the same spacing values. What do you think about making these variables in the `custom` part of theme.json?

scruffian:
For the header and the header patterns, what do you think about using a`<header>` tag? You have to add `"tagName":"header"` to the block....

kjellr:
> Quite a lot of the patterns use the same spacing values. What do you think about making these variables in the custom part of theme.json?

I considered this, but in the interest of portability, I though we should put numerical values in there for now. If we do [make these patterns available in the directory](https://github.com/WordPress/twentytwentytwo/issues/12), we can't guarantee that a users' theme has that variable defined. In any case, if we decide to add that later, it's a quick find and replace. 

> For the header and the header patterns, what do you think about using a<header> tag? You have to add "tagName":"header" to the block....

I thought about this too. We can't guarantee where the user places these patterns, so I wasn't sure if it made sense to mark them as headers? We can _probably_ expect that these are headers since they're labeled as such. I'll update so that it uses the `header` tag.

inc/patterns/header-centered-logo-in-navigation.php
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Header with centered Logo in Navigation block pattern
+ */
+return array(
+	'title'       => __( 'Header with centered Logo in Navigation', 'twentytwentytwo' ),
+	'categories'  => array( 'twentytwentytwo-headers' ),
+	'blockTypes' => array( 'core/template-part/header' ),
+	'content'     => '<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"bottom":"8rem","top":"max(1.25rem, 5vw)"}}},"layout":{"type":"flex","justifyContent":"center"}} -->
+					<div class="wp-block-group alignwide" style="padding-top:max(1.25rem, 5vw);padding-bottom:8rem"><!-- wp:navigation {"itemsJustification":"right","isResponsive":true} -->
+					<!-- wp:page-list {"isNavigationChild":true,"showSubmenuIcon":true,"openSubmenusOnClick":false} /-->
```
jasmussen: I don't think we should use page list here, especially since we're using it twice (it would output the same pages twice over). This pattern, especially in order for the site logo to be centered, is a highly opinionated pattern that should actually come with preset pre-inserted Page Links. They could even come pre-populated with page titles, so all you have to do is click the page item and "wire up" the link. 

When created today, the pattern could/should probably look like this:
pattern-today

Once [#35094](https://github.com/WordPress/gutenberg/pull/35094) lands, we can pre-suggest labels even, making it look like this:
pattern-soon

Yes, highly opinionated, but to me this all ties back to [the conversation about placeholder patterns](https://github.com/WordPress/gutenberg/issues/32373#issuecomment-852201205), where some designs are _created_ around a specific amount of menu items. And so long as it's one alternative among many, hopefully it can inspire, as is part of the purpose. Users will always be able to change labels after the fact, if they like.

kjellr: 👍 Cool, that's updated in ec40e46.

inc/patterns/header-centered-logo.php
```diff
@@ -0,0 +1,26 @@
+<?php
+/**
+ * Header with centered Logo block pattern
+ */
+return array(
+	'title'      => __( 'Header with centered Logo', 'twentytwentytwo' ),
+	'categories' => array( 'twentytwentytwo-headers' ),
+	'blockTypes' => array( 'core/template-part/header' ),
+	'content'    => '<!-- wp:group {"align":"wide","tagName":"header","style":{"spacing":{"padding":{"bottom":"8rem","top":"max(1.25rem, 5vw)"}}}} -->
```
jffng: Most of these header patterns do not have a left / right padding, so when I insert them at the root level, this is how they appear: 

Screen Shot 2021-10-11 at 5 01 06 PM

inc/patterns/header-centered-title-navigation-social.php
```diff
@@ -0,0 +1,30 @@
+<?php
+/**
+ * Centered title with Navigation and Social Links Header block pattern
+ */
+return array(
+	'title'      => __( 'Centered title with Navigation and Social Links Header', 'twentytwentytwo' ),
+	'categories' => array( 'twentytwentytwo-headers' ),
+	'blockTypes' => array( 'core/template-part/header' ),
+	'content'    => '<!-- wp:group {"align":"wide","tagName":"header","style":{"spacing":{"padding":{"bottom":"8rem","top":"max(1.25rem, 5vw)"}}}} -->
```
jffng: With all of these patterns and running Gutenberg trunk, I am getting a header nested inside a header in the markup: 

Screen Shot 2021-10-11 at 5 01 29 PM

inc/patterns/header-centered-logo.php
kjellr: Yeah, I'm unsure of the best way to handle left-right padding for these, but I think leaving it out makes the most sense for now. As I mentioned above: 

> I opted not to include any left/right padding around the patterns themselves. In our templates I've been adding that to [the parent template part](https://github.com/WordPress/twentytwentytwo/blob/af5d31d52c364670a09c8f2a85f77c9a592ca2bd/block-templates/single.html#L1), since it needs to match up with [the padding around the `main` template](https://github.com/WordPress/twentytwentytwo/blob/af5d31d52c364670a09c8f2a85f77c9a592ca2bd/block-templates/single.html#L3) too. Keeping left/right padding out of the patterns seems more versatile than having it in there.

... which I still feel is true. Because for instance, the user will want the same padding around the _entire_ page — the footer, page content, etc.

inc/patterns/header-centered-title-navigation-social.php
kjellr: Oh that's weird. This only happens if you add a `header` block first, and then choose these from the carousel. Then they appear wrapped in a `header`. I guess the solution is to remove the header `tagName` here.

kjellr: Should be all set as of 473937b.

inc/patterns/header-centered-logo.php
mtias: Should the new site padding be used?

kjellr: Yes, it should. 😄 I'll work that into a separate PR.

kjellr: FWIW, I tried this, but it currently leaves us in a worse state than we were in before — the header with the background color isn't possible yet: https://github.com/WordPress/twentytwentytwo/pull/88

jffng (approved):
I think this is a good spot to bring in and we can iterate, since other PRs are relying on it too.

PR #77, opened 2021-10-09
updated readme.txt

Riyadh1734:
Copyright year added


Soean:
Thanks for opening the PR @Riyadh1734, but it's 2021, not 2022, so we can close the PR at the moment.

PR #80, opened 2021-10-11
updated index.php file

Riyadh1734:
added header fields


MaggieCabrera:
I don't think we need this for a block theme

PR #81, opened 2021-10-11
added block image class

Riyadh1734:


scruffian:
Thanks for the PR. Why are you adding the post title and content to the header template?

Riyadh1734:
I think the block templates can have the same header included from a separate header.html template part.

scruffian:
Yes that's right. I'd expect all block templates to use the same header.html template part, but I wouldn't expect the post content to be included in the header - I see the header more for things like the site title, logo and navigation.

Riyadh1734:
what should I do any recommendations?

scruffian:
What are you trying to achieve? I think most of the templates we need are already in place.

Riyadh1734:
If no need for any change that's should be ok  & close it

fabiankaegy:
I am going to close this PR for now based on the last comment here: https://github.com/WordPress/twentytwentytwo/pull/81#issuecomment-940137930

PR #85, opened 2021-10-11
Add footer patterns

### LINKED ISSUE #16: Block Patterns: Add Footer Patterns
Footer Patterns

These mockups are available with more detail in [the theme's Figma file](https://www.figma.com/file/76mfUcaK4QDlrXElk8MK3H/?node-id=758%3A4770).

kjellr:
Closes #16. 

⚠️ _This one relies on some of the block patterns setup from #74, so the PR builds on that for now._

---

This PR adds the 11 footer patterns from [the original mockups](https://www.figma.com/file/76mfUcaK4QDlrXElk8MK3H/Twenty-Twenty-Two?node-id=758%3A4770). 

Thanks @melchoyce for your help building these. 

With a couple small exceptions ([example](https://cloudup.com/cOVD-gazpTj)), these work fine responsively. If those end up being the only issues people see, I'm comfortable merging this and then fixing the individual patterns in followup issues. 

---

## Screenshots

footer-patterns

---

136843375-06af7b26-7f6f-4e8e-9c9f-78f622459996

### DIFF
--- a/inc/patterns/footer-blog.php
+++ b/inc/patterns/footer-blog.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * Blog footer
+ */
+return array(
+	'title'      => __( 'Blog footer', 'twentytwentytwo' ),
+	'categories' => array( 'twentytwentytwo-footers' ),
+	'blockTypes' => array( 'core/template-part/footer' ),
+	'content'    => '<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"8rem","bottom":"8rem"}}},"layout":{"wideSize":"1000px"}} -->
+					<div class="wp-block-group alignfull" style="padding-top:8rem;padding-bottom:8rem"><!-- wp:columns -->
+					<div class="wp-block-columns"><!-- wp:column -->
+					<div class="wp-block-column"><!-- wp:paragraph -->
+					<p>' . esc_html__( 'ABOUT US', 'twentytwentytwo' ) . '</p>
+					<!-- /wp:paragraph -->
+
+					<!-- wp:paragraph -->
+					<p>' . esc_html__( 'We are a rogue collective of bird watchers. We’ve been known to sneak through fences, climb perimeter walls, and generally trespass in order to observe the rarest of birds.', 'twentytwentytwo' ) . '</p>
+					<!-- /wp:paragraph --></div>
+					<!-- /wp:column -->
+
+					<!-- wp:column -->
+					<div class="wp-block-column"><!-- wp:paragraph -->
+					<p>' . esc_html__( 'LATEST POSTS', 'twentytwentytwo' ) . '</p>
+					<!-- /wp:paragraph -->
+
+					<!-- wp:latest-posts /--></div>
+					<!-- /wp:column -->
+
+					<!-- wp:column -->
+					<div class="wp-block-column"><!-- wp:paragraph -->
+					<p>' . esc_html__( 'CATEGORIES', 'twentytwentytwo' ) . '</p>
+					<!-- /wp:paragraph -->
+
+					<!-- wp:categories /--></div>
+					<!-- /wp:column --></div>
+					<!-- /wp:columns -->
+
+					<!-- wp:spacer {"height":50} -->
+					<div style="height:50px" aria-hidden="true" class="wp-block-spacer"></div>
+					<!-- /wp:spacer -->
+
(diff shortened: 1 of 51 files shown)


inc/patterns/footer-dark.php
```diff
@@ -0,0 +1,18 @@
+<?php
+/**
+ * Dark footer wtih title and citation
+ */
+return array(
+	'title'      => __( 'Dark footer wtih title and citation)', 'twentytwentytwo' ),
```
jffng:
```suggestion
	'title'      => __( 'Dark footer with title and citation)', 'twentytwentytwo' ),
```

inc/patterns/footer-title-tagline-social-dark.php
```diff
@@ -0,0 +1,26 @@
+<?php
+/**
+ * Footer with title, tagline, and social links
```
jffng, line 3: This one is not showing up, I think it needs to be added to the array to be registered.

inc/patterns/footer-query-title-citation.php
```diff
@@ -0,0 +1,32 @@
...
+
+					<!-- wp:spacer -->
+					<div style="height:100px" aria-hidden="true" class="wp-block-spacer"></div>
+					<!-- /wp:spacer -->
+
+					<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"4rem","bottom":"4rem"}}},"layout":{"type":"flex","justifyContent":"space-between"}} -->
+					<div class="wp-block-group alignwide" style="padding-top:4rem;padding-bottom:4rem"><!-- wp:site-title /-->
+
+					<!-- wp:paragraph {"align":"right"} -->
+					<p class="has-text-align-right">Proudly Powered by <a href="https://wordpress.org" rel="nofollow">WordPress</a></p>
```
jffng: Should this be escaped + marked for translation?

inc/patterns/footer-query-images-title-citation.php
```diff
@@ -0,0 +1,34 @@
...
+
+					<!-- wp:spacer -->
+					<div style="height:100px" aria-hidden="true" class="wp-block-spacer"></div>
+					<!-- /wp:spacer -->
+
+					<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"4rem","bottom":"4rem"}}},"layout":{"type":"flex","justifyContent":"space-between"}} -->
+					<div class="wp-block-group alignwide" style="padding-top:4rem;padding-bottom:4rem"><!-- wp:site-title /-->
+
+					<!-- wp:paragraph {"align":"right"} -->
+					<p class="has-text-align-right">Proudly Powered by <a href="https://wordpress.org" rel="nofollow">WordPress</a></p>
```
jffng: Should this be escaped + marked for translation?

inc/patterns/footer-blog.php
```diff
@@ -0,0 +1,50 @@
...
+ */
+return array(
+	'title'      => __( 'Blog footer', 'twentytwentytwo' ),
+	'categories' => array( 'twentytwentytwo-footers' ),
+	'blockTypes' => array( 'core/template-part/footer' ),
+	'content'    => '<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"8rem","bottom":"8rem"}}},"layout":{"wideSize":"1000px"}} -->
+					<div class="wp-block-group alignfull" style="padding-top:8rem;padding-bottom:8rem"><!-- wp:columns -->
+					<div class="wp-block-columns"><!-- wp:column -->
+					<div class="wp-block-column"><!-- wp:paragraph -->
+					<p>' . esc_html__( 'ABOUT US', 'twentytwentytwo' ) . '</p>
```
jffng, line 13: Can these labels be Sentence cased and then `"textTransform": uppercase;`d?

jffng (commented):
Looking good! 

One general comment, most (all?) of these don't have any left / right padding so they butt up against the edges on mobile:

Screen Shot 2021-10-11 at 5 56 36 PM

inc/patterns/footer-blog.php
kjellr: I think we should leave them as is for a couple reasons: 

- The paragraph block does not have a visible textTransform control. And entering that manually on the block doesn't seem to work in my testing. 
- I could switch to a heading block here, but I want these to be set in the system font, and a font choice isn't available for the heading block. (I think this was originally to limit the user to sticking with the theme's styles, but I find it pretty limiting myself... I might open a Gutenberg issue to reconsider that). 

For now, I think this is the simplest approach, and will make the most sense to users editing this one.

jffng (approved):

Issue #5, opened 2021-10-04
Ensure full-width items within post content go full-width


kjellr:
_Flagging this issue from https://github.com/WordPress/twentytwentytwo/pull/4#discussion_r721786880 to follow up later:_ 

Due to some left/right padding rules, `alignfull` items within posts and pages do not actually extend to the full width of the page:

<img width="1272" alt="Screen Shot 2021-10-04 at 7 40 10 PM" src="https://user-images.githubusercontent.com/1202812/135938627-d5c08a48-9164-4ae2-b3a3-497760e674e2.png">

We should (hopefully) find a way to allow for full-width items to go all the way to the screen edges, while still retaining a similar measure of left/right padding for the header and other non-full-width items.

pattonwebz:
The most common technique I have seen used to do this is with CSS similar to this:

```
.alignfull {
	margin: 2rem calc(50% - 50vw);
	max-width: 100vw;
	padding: 1rem 2rem;
	width: 100vw;
}
```

That has some implications on chrome where it can cause a horizontal scroll bar due to chrome (possibly all webkit browsers) not accounting for the vertical scrollbar when it calculates what 100vw is. I think it's worth exploring if this will be suitable and if the kinds with chrome can be worked out.

kjellr:
Closing to consolidate this issue into https://github.com/WordPress/twentytwentytwo/issues/234.

PR #88, opened 2021-10-12
Try using global site padding

kjellr:
This PR is an attempt to use the new global site padding `theme.json` entry as introduced in https://github.com/WordPress/gutenberg/pull/35241.

For most templates, things look identical to what we had before: 

Screen Shot 2021-10-12 at 8 00 48 AM

... however there's one major issue: the padding also applies to items that are set to full-width. This means we cannot have the full-width dark header: 

Screen Shot 2021-10-12 at 7 56 36 AM

Similarly, full-width items in the post content do not extend to the edges of the screen ([they don't do that with our current approach either](https://github.com/WordPress/twentytwentytwo/issues/5)). Not sure the best way around this — but I'd expect that if an item is specifically set to appear full-width, Gutenberg should let it be full-width?

### DIFF
--- a/block-template-parts/header-large-dark.html
+++ b/block-template-parts/header-large-dark.html
@@ -1,5 +1,5 @@
-<!-- wp:group {"style":{"elements":{"link":{"color":{"text":"var:preset|color|background"}}},"spacing":{"padding":{"top":"0px","right":"max(1.25rem, 5vw)","bottom":"max(1.25rem, 5vw)","left":"max(1.25rem, 5vw)"}}},"backgroundColor":"foreground","textColor":"background","layout":{"inherit":true}} -->
-<div class="wp-block-group has-background-color has-foreground-background-color has-text-color has-background has-link-color" style="padding-top:0px;padding-right:max(1.25rem, 5vw);padding-bottom:max(1.25rem, 5vw);padding-left:max(1.25rem, 5vw)"><!-- wp:template-part {"slug":"header","tagName":"header","layout":{"inherit":true},"align":"wide"} /-->
+<!-- wp:group {"style":{"elements":{"link":{"color":{"text":"var:preset|color|background"}}},"spacing":{"padding":{"top":"0px","bottom":"max(1.25rem, 5vw)"}}},"backgroundColor":"foreground","textColor":"background","layout":{"inherit":true}} -->
+<div class="wp-block-group has-background-color has-foreground-background-color has-text-color has-background has-link-color" style="padding-top:0px;padding-bottom:max(1.25rem, 5vw)"><!-- wp:template-part {"slug":"header","tagName":"header","layout":{"inherit":true},"align":"wide"} /-->
 
 <!-- wp:columns {"align":"wide"} -->
 <div class="wp-block-columns alignwide"><!-- wp:column -->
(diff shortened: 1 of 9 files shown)


block-template-parts/header-large-dark.html
```diff
@@ -1,5 +1,5 @@
-<!-- wp:group {"style":{"elements":{"link":{"color":{"text":"var:preset|color|background"}}},"spacing":{"padding":{"top":"0px","right":"max(1.25rem, 5vw)","bottom":"max(1.25rem, 5vw)","left":"max(1.25rem, 5vw)"}}},"backgroundColor":"foreground","textColor":"background","layout":{"inherit":true}} -->
-<div class="wp-block-group has-background-color has-foreground-background-color has-text-color has-background has-link-color" style="padding-top:0px;padding-right:max(1.25rem, 5vw);padding-bottom:max(1.25rem, 5vw);padding-left:max(1.25rem, 5vw)"><!-- wp:template-part {"slug":"header","tagName":"header","layout":{"inherit":true},"align":"wide"} /-->
+<!-- wp:group {"style":{"elements":{"link":{"color":{"text":"var:preset|color|background"}}},"spacing":{"padding":{"top":"0px","bottom":"max(1.25rem, 5vw)"}}},"backgroundColor":"foreground","textColor":"background","layout":{"inherit":true}} -->
```
kjellr, line 1: Also: I removed the left/right padding from here, but that defaults us to the default "Group block with background color" padding. It would be cool if that arbitrary value was automatically replaced with the global site padding value when one exists.

kjellr:
I had some followup discussions with @jasmussen about this, and it seems like solving for that full-width header while also using global site padding gets us into the weeds pretty quickly. Either we'd need to add some custom CSS with negative margins (I thought we'd left that world when we introduced Layout in Gutenberg! 😩), or we'd need to add CSS like that in Gutenberg to handle this there. For the latter though, it's difficult to figure out whether or not a user actually _does_ want full-width items to go outside of the padding. After all, padding is generally meant to apply to all children. 

So in general, since this theme relies on a core element (that homepage header) being full-width, I think this global site padding is probably not a tool that we should be using.

jffng:
> I think this global site padding is probably not a tool that we should be using.

I agree, at least not in its current iteration. We need the ability for blocks at the root level (and ideally also blocks within pages and posts per #5) to actually be full width / extend to the edge of the page.

jasmussen:
> We need the ability for blocks at the root level (and ideally also blocks within pages and posts per #5) to actually be full width / extend to the edge of the page.

The block editor can absorb some complexity, and when we can, it's great. But ultimately we are bound by needing to output CSS in the end, and so zooming way out, the problem here remains one of deciding where padding is applied. If it's at the root level, how do we then stretch items over that padding? If we can build a nice codepen, we can work from there.

jffng:
> how do we then stretch items over that padding?

What are your thoughts here? In Blockbase we [use negative margins](https://github.com/Automattic/themes/blob/trunk/blockbase/sass/base/_alignment.scss#L8-L9), and have had to refactor this bit many times to account for various use and edge cases. I'm happy to explore some other ideas.

jasmussen:
My first instinct was negative margins as well. One challenge there is to "automate" them, as the negative margin has to match the positive padding. That can be done with a CSS variable and a little calc thrown in for good measure, but wiring that up to a good experience in global styles can be gnarly.

kjellr:
I'm going to close this one for now, while we discuss possible solutions in https://github.com/WordPress/gutenberg/issues/35607.

Issue #21, opened 2021-10-05
Set up 2022.wordpress.net


jffng:
We should set up a development site + a github action that deploys `trunk` of this theme. This is probably separate from 2022.wordpress.net?

kraftbj:
It could be 2022.wordpress.net. The action would need to push up via SFTP/SSH and we'd need to create a key for it. Previous sites, I manually updated it ~weekly during dev.

Happy to own this one.

kjellr:
Thanks, @kraftbj!

pbking:
@kraftbj I opened #94 for an alternate "workshop" location where we can have an uh... less appealing and more edge-casey site than 2022.wordpress.net while we bang out some kinks.  

Perhaps that action could be augmented to also include 2022.wordpress.net for a location with cleaner content?

kraftbj:
I like that idea @pbking. Either augment or have a second one that does the same thing. I've requested the site and asking the Systems folks if they prefer a SSH keypair or user/pass for a dedicated SFTP user for this purpose. Thank you!

kraftbj:
@kjellr I've setup an admin user for you on the site. You can password recover with your automattic e-mail address (check spam; the wordpress.net servers aren't well-liked).

I've setup Akismet and added a comped account. Comments are all moderated, anyhow, but no need to keep spam around.

kjellr:
Thanks for the help, @kraftbj! I'll keep this issue open while I populate it with demo content later today, but I'll reassign it to myself.

PR #94, opened 2021-10-12
Create staging_workshop.yml

pbking:
Adding a deploy action on changes to /trunk for https://twentytwentytwo-trunk.mystagingwebsite.com/

**Description**

This deploys the content of the repo to the /themes folder in the noted site.  The expectation is that this would be the "workshop" where there will be many examples of different types of content and edge conditions.

The "starting content" is the [theme-test-data](https://github.com/WPTT/theme-test-data).

A separate site (likely 2022.wordpress.net) should also be setup to be a more "public face" of the theme with content better designed to demonstrate the strengths of the theme and properly achieve the design targets.

`STAGING_USER` and `STAGING_PASSWORD` values will need to be stored as secrets in order for this action to work.

Addresses #21 (though I still think 2022.wordpress.net would benefit from an automated deployment alongside this)

### DIFF
--- a/.github/workflows/staging_workshop.yml
+++ b/.github/workflows/staging_workshop.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Deploy-Theme
         uses: SamKirkland/FTP-Deploy-Action@3.1.1
         with:
-          ftp-server: sftp://sftp.pressable.com:22/htdocs/wp-content/themes/
+          ftp-server: sftp://sftp.pressable.com:22/htdocs/wp-content/themes/twentytwentytwo/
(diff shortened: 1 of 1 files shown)


.github/workflows/staging_workshop.yml
```diff
@@ -0,0 +1,24 @@
+name: TwentyTwentyTwo Staging Workshop Deploy
```
jffng:
```suggestion
name: Deploy to Test Site
```

jffng (approved):
I added the respective secrets to the repository, so we should be all set.

Left one small comment about the workflow name, thank you for setting this up @pbking !

PR #96, opened 2021-10-13
Simplify dark header template and pattern.

kjellr:
This PR simplifies the block setup for the Dark header template and pattern (used on the  homepage). 

- It relies on the recently-merged https://github.com/WordPress/gutenberg/pull/35589 to make the site tagline wide-width. Previously, the tagline was wrapped in a columns block that added a bunch of extra unnecessary margins. 
- Since that weird extra margin is gone, this PR also re-adjusts some of the paddings/margin to make things look a little bit more even. 

When testing, you'll need to be running the latest Gutenberg `trunk`. (If it works, I think it's fine to merge this in now, even though that change hasn't made it to the plugin yet). 

Before|After
---|---
Screen Shot 2021-10-13 at 09 01 33|Screen Shot 2021-10-13 at 09 01 06

### DIFF
--- a/inc/patterns/header-large-dark.php
+++ b/inc/patterns/header-large-dark.php
@@ -14,7 +14,7 @@ return array(
 					<!-- wp:site-title {"style":{"typography":{"fontStyle":"italic","fontWeight":"400"}}} /--></div>
 					<!-- /wp:group -->
 
-					<!-- wp:navigation {"itemsJustification":"right"} -->
+					<!-- wp:navigation {"itemsJustification":"right","isResponsive":true} -->
 					<!-- wp:page-list {"isNavigationChild":true,"showSubmenuIcon":true,"openSubmenusOnClick":false} /-->
 					<!-- /wp:navigation --></div>
 					<!-- /wp:group -->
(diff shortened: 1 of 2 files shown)


inc/patterns/header-large-dark.php
```diff
@@ -6,24 +6,19 @@
...
 					<div class="wp-block-group alignwide" style="padding-top:max(1.25rem, 5vw);padding-bottom:8rem"><!-- wp:group {"layout":{"type":"flex"}} -->
-					<div class="wp-block-group">
-					<!-- wp:site-logo {"width":64} /-->
+					<div class="wp-block-group"><!-- wp:site-logo {"width":64} /-->
 
 					<!-- wp:site-title {"style":{"typography":{"fontStyle":"italic","fontWeight":"400"}}} /--></div>
 					<!-- /wp:group -->
 
-					<!-- wp:navigation {"itemsJustification":"right","isResponsive":true} -->
+					<!-- wp:navigation {"itemsJustification":"right"} -->
```
jffng: Did you mean to remove the `isResponsive` attribute?

jffng (approved):
Just one comment about responsive menu on the nav, otherwise this LGTM.

inc/patterns/header-large-dark.php
kjellr: I did not! Weird. I'll put it back in.

kjellr:
```suggestion
					<!-- wp:navigation {"itemsJustification":"right","isResponsive":true} -->
```

Issue #89, opened 2021-10-12
Header Patterns: Followup Fixes


kjellr:
Following up from #74, there are a few header patterns with unfortunate responsive behavior. I'm creating this issue so that we can continue to track the issues: 

## Header with Tagline

- [ ] This wraps poorly on small screens. ([Screenshot](https://cloudup.com/cxBvFm2Oj3V))

## Text-only Header with Tagline

- [ ] This wraps poorly on small screens. ([Screenshot](https://cloudup.com/cz_z0CveuCt))

## Text-only Header with Tagline and all-caps Title

- [ ] This wraps poorly on small screens. ([Screenshot](https://cloudup.com/c4mv-0BEeLj))

## Header with centered Logo

- [ ] This wraps poorly on small screens. [(GIF)](https://cloudup.com/ciCs3nzlqFN)

## Header with centered Logo in Navigation

- [ ] The site logo disappears on smaller screens. ([Screenshot](https://cloudup.com/c-krv3vU4u7))
- [ ] The layout of the expanded menu looks weird. ([Screenshot](https://cloudup.com/cBDKpdI0Fbz))
- [ ] Depending on the number/length of menu items, the logo is not necessarily located in the exact center. ([Screenshot](https://cloudup.com/cZzkpra6jFl))

## Centered title with Navigation and Social Links Header

- [ ] The items wrap oddly on small screens. ([Screenshot](https://cloudup.com/c91b-Oh1-oD))
- [ ] The navigation items start stacking in a weird way on medium breakpoints. ([Screenshot](https://cloudup.com/cq9BAkOnkXk))

## Title and Button Header

- [x] This one is supposed to have the hamburger icon visible by default on all breakpoints. Requires a resolution to https://github.com/WordPress/gutenberg/issues/34783.

kjellr:
>  This wraps poorly on small screens

^ Regarding all of these bugs: Now that we've gone through all the trouble of getting https://github.com/WordPress/gutenberg/pull/36003 merged, I'm finding more use cases where these look better without that control. For example, take this example with a bunch of menu items and a long tagline.

Here is is normally: 

Desktop|Mobile
---|---
[...]

PR #97, opened 2021-10-13
Update Title and Button Header pattern

kjellr:
Partially addresses #89. 
Relies on https://github.com/WordPress/gutenberg/pull/35568, and can be merged as soon as that lands.

This PR implements the "always on" responsive option for the "Title and Button" header pattern. This syncs it up with the way it was originally designed. I've also moved it up in the list, since I think it's worth highlighting this more prominently. 

Before:

Screen Shot 2021-10-13 at 9 15 55 AM

After:
Screen Shot 2021-10-13 at 9 12 56 AM


jffng (approved):
Works, thanks!

PR #98, opened 2021-10-13
Fix WordPress link internationalization in footer patterns

audrasjb:
**Description**

The "WordPress" link in footer patterns needs to be translatable, since Locales will replace the link with the link to the Rosetta website.

**Screenshots**

Not applicable.

**Testing Instructions** 

Not easily testable, but I tested to insert the edited patterns and it works fine :)

### DIFF
--- a/inc/patterns/footer-blog.php
+++ b/inc/patterns/footer-blog.php
@@ -43,7 +43,7 @@ return array(
 					<div class="wp-block-group alignwide" style="padding-top:4rem;padding-bottom:4rem"><!-- wp:site-title /-->
 
 					<!-- wp:paragraph {"align":"right"} -->
-					<p class="has-text-align-right">' . esc_html__( 'Proudly powered by ', 'twentytwentyone' ) . '<a href="' . esc_url( 'https://wordpress.org' ) . '" rel="nofollow">WordPress</a></p>
+					<p class="has-text-align-right">' . esc_html__( 'Proudly powered by ', 'twentytwentyone' ) . '<a href="' . esc_url( __( 'https://wordpress.org', 'twentytwentytwo' ) ) . '" rel="nofollow">WordPress</a></p>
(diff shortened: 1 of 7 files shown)


inc/patterns/footer-blog.php
```diff
@@ -43,7 +43,7 @@
 					<div class="wp-block-group alignwide" style="padding-top:4rem;padding-bottom:4rem"><!-- wp:site-title /-->
 
 					<!-- wp:paragraph {"align":"right"} -->
-					<p class="has-text-align-right">' . esc_html__( 'Proudly powered by ', 'twentytwentyone' ) . '<a href="' . esc_url( 'https://wordpress.org' ) . '" rel="nofollow">WordPress</a></p>
+					<p class="has-text-align-right">' . esc_html__( 'Proudly powered by ', 'twentytwentyone' ) . '<a href="' . esc_url( __( 'https://wordpress.org', 'twentytwentytwo' ) ) . '" rel="nofollow">WordPress</a></p>
```
jffng, line 46: [twentytwentyone](https://github.com/WordPress/twentytwentyone/blob/trunk/footer.php#L61) uses `esc_attr__` here instead, is there a meaningful difference between this approach?

kjellr: `esc_url()` feels more appropriate to me since it is a URL. We want to make sure it stays a URL after all. 🙂

Looks like [Twenty Nineteen](https://github.com/WordPress/twentynineteen/blob/b604f127c2cae10bd48bbbec0fbbbff2cd31f957/footer.php#L25) and [Twenty Twenty](https://github.com/WordPress/twentytwenty/blob/6d0a5240af108d02f58ec797fd49b32458a4c698/footer.php#L32) use `esc_url()`, so I think we're good.

kjellr (approved):
Thanks for the fix, @audrasjb!

PR #104, opened 2021-10-13
Add page block patterns

### LINKED ISSUE #19: Block Patterns: Add Page Layout Patterns
These patterns are designed as sort of "posters" for the web, intended to be used as the entirety of the page content. 

Frame 2

These mockups are available with more detail in [the theme's Figma file](https://www.figma.com/file/76mfUcaK4QDlrXElk8MK3H/?node-id=758%3A4770).

I'm including `[Status] Needs Design` because I intend to add more of these.

MaggieCabrera:
**Description**

Closes https://github.com/WordPress/twentytwentytwo/issues/19

This PR creates a blank page template that only adds the post-content block and the 3 page layout patterns from https://github.com/WordPress/twentytwentytwo/issues/19

**Testing Instructions** 

1. Create a new page
2. Select the Blank page template on the sidebar
3. Insert one of the new patterns under the `Twenty Twenty-Two Page layouts` category and check that everything works and looks as intended


inc/patterns/page-layout-image-and-text.php
```diff
@@ -0,0 +1,27 @@
...
+ */
+return array(
+	'title'      => __( 'Page layout with image and text', 'twentytwentytwo' ),
+	'categories' => array( 'twentytwentytwo-pages' ),
+	'content'    => '<!-- wp:heading {"align":"wide","style":{"typography":{"fontSize":"120px","lineHeight":"1.15","fontWeight":"300"}}} -->
+	<h2 class="alignwide" style="font-size:120px;font-weight:300;line-height:1.15">' . wp_kses_post( __('<em>Watching Birds </em><br><em>in the Garden</em>', 'twentytwentytwo') ) . '</h2>
+	<!-- /wp:heading -->
+	
+	<!-- wp:image {"align":"full","style":{"color":{}}} -->
+	<figure class="wp-block-image alignfull"><img src="' . esc_url( get_template_directory_uri() ) . '/assets/img/patterns/watching-birds.png" alt="' . esc_attr__( 'TBD', 'twentytwentytwo' ) . '"/></figure>
```
MaggieCabrera: @melchoyce @kjellr I'm missing the alt text for the 2 images on these patterns.

jffng (commented):
Thanks for picking these up @MaggieCabrera !

When used with the `blank` template, it looks like each of these patterns could use some top and bottom padding:

Screen Shot 2021-10-13 at 1 58 19 PM

inc/patterns/page-layout-image-text-and-video.php
```diff
@@ -0,0 +1,59 @@
+<?php
+/**
+ * Page layout with image, text and video.
```
jffng, line 3: There is quite a bit of space between blocks on the front-end with this pattern:
Screen Shot 2021-10-13 at 2 01 11 PM

melchoyce: There's some invisible text there — it needs a green background somehow.

melchoyce:
> Thanks for picking these up @MaggieCabrera !
> 
> When used with the `blank` template, it looks like each of these patterns could use some top and bottom padding

In my test site, I made my own blank template with a 100px spacer above and below the page content, fwiw.

kjellr:
> In my test site, I made my own blank template with a 100px spacer above and below the page content, fwiw.

I think we can just add something like that to each of these patterns. Maybe borrow the padding values used around our header templates?

inc/patterns/page-layout-image-text-and-video.php
kjellr: We should wrap the whole thing in a green group block.

MaggieCabrera: I'll get to it!

MaggieCabrera: it should look much better now:
FireShot Capture 018 - Test blank – tt2 - tt2 local

inc/patterns/page-layout-image-text-and-video.php
```diff
@@ -0,0 +1,64 @@
...
+ */
+return array(
+	'title'      => __( 'Page layout with image, text and video', 'twentytwentytwo' ),
+	'categories' => array( 'twentytwentytwo-pages' ),
+	'content'    => '<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"8rem","bottom":"8rem"}}},"backgroundColor":"primary","textColor":"background"} -->
+	<div class="wp-block-group alignfull has-background-color has-primary-background-color has-text-color has-background" style="padding-top:8rem;padding-bottom:8rem"><!-- wp:group {"align":"wide","layout":{"inherit":true}} -->
```
MaggieCabrera: do we want to use the hex values instead of classes here for better cross theme support? I know there's discussions about this over at the GB repo but I don't know if that will come to something before TT2 launches

MaggieCabrera:
I added a group to all of these patterns and the 8rem vertical padding that the header and footer of the theme applies is applied here too:

Screenshot 2021-10-14 at 09-31-39 Test blank – tt2


Screenshot 2021-10-14 at 09-33-37 Test blank – tt2

MaggieCabrera:
this should be ready for another review, pending my comments about the colors and the missing alt texts, I think these patterns should be ready now.

inc/patterns/page-layout-image-and-text.php
```diff
@@ -0,0 +1,29 @@
+<?php
+/**
+ * Page layout with image and text.
+ */
+return array(
+	'title'      => __( 'Page layout with image and text', 'twentytwentytwo' ),
+	'categories' => array( 'twentytwentytwo-pages' ),
+	'content'    => '<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"8rem","bottom":"8rem"}}}} -->
+	<div class="wp-block-group alignwide" style="padding-top:8rem;padding-bottom:8rem"><!-- wp:heading {"align":"wide","style":{"typography":{"fontSize":"120px","lineHeight":"1.15","fontWeight":"300"}}} -->
+	<h2 class="alignwide" style="font-size:120px;font-weight:300;line-height:1.15">' . wp_kses_post( __( '<em>Watching Birds </em><br><em>in the Garden</em>', 'twentytwentytwo' ) ) . '</h2>
```
kjellr: All of these large text values should actually be responsive `clamp()` values that use `rem` instead of pixels. See what we're doing for all our presets for example: 

https://github.com/WordPress/twentytwentytwo/blob/c34addcc6f8eae0be67e5015fc11411214d57bd3/theme.json#L39

I can take a look in a couple hours and advise on some replacement values. 👍

kjellr: This headline can be `clamp(4rem, 8vw, 7.5rem)` instead of `120px`.

inc/patterns/page-layout-image-text-and-video.php
kjellr: Not yet. If we move this over to the pattern directory, we can swap these out with hex values. But right now there's value in using the theme variables here since it'll continue to update the colors in sync with Global Styles.

inc/patterns/page-layout-image-text-and-video.php
```diff
@@ -0,0 +1,64 @@
...
+	<div class="wp-block-group alignwide">
+	<!-- wp:heading {"level":1,"align":"wide","style":{"typography":{"fontSize":"72px"}}} -->
```
kjellr: This font size should can be `clamp(3rem, 6vw, 4.5rem)`.

inc/patterns/page-layout-image-text-and-video.php
```diff
@@ -0,0 +1,64 @@
...
+	<!-- /wp:heading -->
+	
+	<!-- wp:spacer {"height":50} -->
+	<div style="height:50px" aria-hidden="true" class="wp-block-spacer"></div>
+	<!-- /wp:spacer -->
+	
+	<!-- wp:cover {"url":"' . esc_url( get_template_directory_uri() ) . '/assets/img/patterns/warble.png","focalPoint":{"x":"0.89","y":"0.38"},"minHeight":900,"contentPosition":"top left","align":"wide"} -->
+	<div class="wp-block-cover alignwide has-custom-content-position is-position-top-left" style="min-height:900px"><span aria-hidden="true" class="has-background-dim-100 wp-block-cover__gradient-background has-background-dim"></span><img class="wp-block-cover__image-background" alt="' . esc_attr__( 'TBD', 'twentytwentytwo' ) . '" src="' . esc_url( get_template_directory_uri() ) . '/assets/img/patterns/warble.png" style="object-position:89% 38%" data-object-fit="cover" data-object-position="89% 38%"/><div class="wp-block-cover__inner-container"><!-- wp:heading {"style":{"typography":{"fontSize":"32px"}}} -->
+	<h2 style="font-size:32px">' . esc_html__( 'Screening', 'twentytwentytwo' ) . '</h2>
```
kjellr: Instead of a custom pixel font size, we should make this (and the other similar heading down by the video) use the theme's `Large` preset size.

inc/patterns/page-layout-two-columns.php
```diff
@@ -0,0 +1,73 @@
+<?php
+/**
+ * Page layout with two columns.
+ */
+return array(
+	'title'      => __( 'Page layout with two columns', 'twentytwentytwo' ),
+	'categories' => array( 'twentytwentytwo-pages' ),
+	'content'    => '<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"8rem","bottom":"8rem"}}}} -->
+	<div class="wp-block-group alignwide" style="padding-top:8rem;padding-bottom:8rem"><!-- wp:heading {"level":1,"align":"wide","style":{"typography":{"fontSize":"200px","lineHeight":"1","fontWeight":"200"}}} -->
+	<h1 class="alignwide" style="font-size:200px;font-weight:200;line-height:1">' . wp_kses_post( __( '<em>Goldfinch </em><br><em>&amp; Sparrow</em>', 'twentytwentytwo' ) ) . '</h1>
```
kjellr: The font size here can be `clamp(4rem, 15vw, 12.5rem)`.

kjellr:
I left some font size updates above. One other small comment: Some of these bump up against the edges of the screen: 

Screen Shot 2021-10-14 at 7 58 32 AM

We should apply the same (admittedly weird) left + right padding values to these that we use for the main body content in our templates: 

https://github.com/WordPress/twentytwentytwo/blob/1a98ab4969149d55e96930b5cd7d2e35b16fea39/block-templates/index.html#L3

(this should apply to the interior left/right padding of the green group container as well.)

inc/block-patterns.php
```diff
@@ -20,6 +20,10 @@ function twentytwentytwo_register_block_patterns() {
 				'twentytwentytwo-query',
 				array( 'label' => __( 'Twenty Twenty-Two Posts', 'twentytwentytwo' ) )
 			);
+			register_block_pattern_category(
+				'twentytwentytwo-pages',
+				array( 'label' => __( 'Twenty Twenty-Two Page layouts', 'twentytwentytwo' ) )
```
kjellr:
```suggestion
				array( 'label' => __( 'Twenty Twenty-Two Page Layouts', 'twentytwentytwo' ) )
```

Small note: This feels weird lowercase since no other category has a lowercase word. 😄 Let's just make it uppercase for now. 

Thanks!

inc/patterns/page-layout-image-text-and-video.php
MaggieCabrera: Makes sense to me!

MaggieCabrera:
I added the lateral paddings and changed the font sizes as suggested

kjellr:
Thanks, Maggie! I made a few adjustments: 

- Added alt text where it was missing. 
- Changed the padding around so that the full-width bird images could go full-width.
- Switched away from using a Cover block for the green one, for better responsive behavior. 
- Indented to match the format of existing patterns. 
- Compressed/trimmed the video so it's 3.6mb (it was over 20mb before!)
- Changed the category name from "Layout Pages" to simply "Pages". 
- Added more complete image credits.

This should be in a good spot for another review!

jffng (approved):
👏 So nice.

PR #105, opened 2021-10-13
Update Dark Header to reuse the Header template

kjellr:
Followup from #96. This re-integrates the workaround from https://github.com/WordPress/twentytwentytwo/pull/55. 

I missed that originally, and didn't mean to remove it here. 👍


jffng (approved):
Whoops, thanks for the follow-up.

PR #115, opened 2021-10-14
Tidy up post meta

kjellr:
This PR does some minor cleanup to the post meta. It just puts everything on one line. 

It's not what we had originally mocked up, but I think is more versatile. I may re-think it later on but in the meantime this is cleaner than what we have. 

**Before:** 
Screen Shot 2021-10-14 at 3 58 09 PM

**After:**
Screen Shot 2021-10-14 at 4 02 57 PM


jffng (approved):
I like the simplification.

Issue #111, opened 2021-10-14
Remove unnecessary theme supports from functions.php 


kjellr:
https://github.com/WordPress/gutenberg/pull/35593 declares the following by default for all block themes: 

```
add_theme_support( 'post-thumbnails' );
add_theme_support( 'responsive-embeds' );
add_theme_support( 'editor-styles' );
add_theme_support(
	'html5',
	array(
		'style',
		'script',
	)
);
add_theme_support( 'automatic-feed-links' );
add_filter( 'should_load_separate_core_block_assets', '__return_true' );
```

These no longer need to be included in Twenty Twenty-Two's file.

juricav:
@kjellr hi, I've made the changes in the file

PR #125, opened 2021-10-16
Add Support for Post Thumbnails / Featured Images

aaronjorbin:
The theme currently uses Post Thumbnails in templates, but without explicitly declaring support for it, it is impossible to edit them or add new ones.

**Description**

In [multiple places the theme uses the featured image](https://github.com/WordPress/twentytwentytwo/search?q=wp%3Apost-featured-image), let's make it possible to also edit them


Ismail-elkorchi:
Post Thumbnails / Featured Images support is enabled by default in all block themes. See https://github.com/WordPress/twentytwentytwo/issues/111

kjellr:
That's right, it's declared by default for block themes now. I'm going to close, but thanks for the PR!

PR #127, opened 2021-10-18
Add missing comments block to page template

audrasjb:
**Description**

Twenty Twenty-Two must support comments on Page post type, if enabled on a specific page using the "discussion" settings:
Capture d’écran 2021-10-18 à 10 26 39

We simply need to add the comment block template to the page.html template.

**Screenshots**

Before the PR: a Page with comment option enabled didn't show any comment/comment form.
Capture d’écran 2021-10-18 à 10 27 02

After the PR: comments/comment form is displayed when the option is enabled.
Capture d’écran 2021-10-18 à 10 35 52

**Testing Instructions** 

1. Enable comments on a Page using the "Discussion" settings of this specific page.
2. On front-end, check wether the comments list/comment form is displayed.

PR #131, opened 2021-10-18
Hide "Comments Closed" message

jffng:
**Description**

This PR is an alternative to #127 per some discussion in this trac ticket https://core.trac.wordpress.org/ticket/54283 (thanks @audrasjb for opening). It depends on https://github.com/WordPress/gutenberg/pull/35743

**Testing Instructions** 

1. Check out the above Gutenberg PR and build it.
2. Go to a post or page whose comments are disallowed
3. Verify the message is hidden


kjellr:
Assuming this attribute is needed, this could probably be worked into #127 (or vice versa? That PR now includes some spacing stuff that will be needed either way).

jffng:
> Assuming this attribute is needed, this could probably be worked into #127 (or vice versa? That PR now includes some spacing stuff that will be needed either way).

Yes, can work this into #127. But your question made me think, do we even need the attribute?

kjellr:
Yeah, I don't think we do. That message should just be hidden when there are no comments + comments are closed.

jffng:
I updated the Gutenberg PR to do that, going to close this one in favor of #127.

PR #134, opened 2021-10-19
Set default text size for buttons

kjellr:
[As per the mockups](https://www.figma.com/file/76mfUcaK4QDlrXElk8MK3H/?node-id=729%3A4866), the default text size for buttons should be our `normal` font size. They're currently actually a bit larger than that. 

Before|After
---|---
Screen Shot 2021-10-19 at 10 19 56 AM|Screen Shot 2021-10-19 at 10 19 44 AM


jffng (approved):
Ah this was bothering me, thanks.

PR #136, opened 2021-10-19
Update post, page, and index templates

kjellr:
I've been [tidying up many of our templates in Figma](https://www.figma.com/file/76mfUcaK4QDlrXElk8MK3H/?node-id=729%3A4819), and this is an attempt to put those new templates into action. (Apologies in advance for the larger PR, but one small change here and there led to another). A few general notes before I dive into specifics: 

- Some of the new templates rely on applying margin or padding to the post titles or featured images. These are reliant on two Gutenberg PRs landing: https://github.com/WordPress/gutenberg/pull/35775 and https://github.com/WordPress/gutenberg/pull/35772
- I added a `customTemplates` section to our `theme.json`. We should've had this there originally, for the homepage and blank templates. 
- In a couple spots, I implemented responsive height values. This ensures that [large empty spaces scale down proportionally for mobile](https://cloudup.com/cAPx_v6Q0Xt). I'd like to do this more, but some areas rely on the Spacer block for layout, and that only accepts pixel values at the moment: https://github.com/WordPress/gutenberg/issues/22956
- Some of the spacing values we were using were getting out of control and repetitive, so I opted to define some custom values. Whenever I used them in templates, I declared the CSS variables with fallbacks (e.g. `var(--variable-name, 1rem)`), just in case someone kept these templates and changed themes or something. 

## Index Template

This PR switches out the base index template with a fresh one that's more flexible. Previously, the default template was a grid, which looked pretty broken when someone had a mix of posts with featured images and without. This new layout is a bit more of a traditional blog layout, which scales a little better that way. 

- The PR makes sure that the `index.html` inherits the main query, and it also adds a Query Title block. Combined, that ensures that this same template can work for archive pages. 
- The old template is still available as a Query pattern, with the new title "Grid of posts". I did some light renaming of other query grid patterns to match. 

Blog index:

Before|After
---|---
Screen Shot 2021-10-19 at 15 43 21|Screen Shot 2021-10-19 at 15 42 43

Archive page: 

Before|After
---|---
Screen Shot 2021-10-19 at 16 05 40|Screen Shot 2021-10-19 at 16 05 31

## Single & Page Templates

In the Figma file, I did some explorations around how to standardize spacing, eliminate the "boxy" feel we were getting from tons of dividers, and to better handle situations when there was no featured image. This PR implements those new changes. 

Posts:

Before|After
---|---
Screen Shot 2021-10-19 at 15 49 52|Screen Shot 2021-10-19 at 15 48 47

Pages: 

Before|After
---|---
Screen Shot 2021-10-19 at 15 51 43|Screen Shot 2021-10-19 at 15 52 27

## Custom Templates

As introduced in the Figma file, I added a couple custom templates as well. These look identical to the normal ones, except that they do not have any separators. When testing these templates out, you may notice that all templates appear for all post types, even though they have specific `postTypes` defined. There's a Gutenberg bug already filed for this here: https://github.com/WordPress/gutenberg/issues/31704


kjellr:
I may end up making some small design changes to the `index.html` template, but the rest should be fair game for review.

audrasjb:
FWIW, those changes looks relevant to me 👍

kjellr:
> I may end up making some small design changes to the index.html template, but the rest should be fair game for review.

With eda696b, this is in a decent spot for now, and I can iterate later on if need be. The whole PR should be ready to review!

block-templates/index.html
```diff
@@ -1,29 +1,45 @@
...
 
-<!-- wp:group {"style":{"spacing":{"padding":{"right":"max(1.25rem, 5vw)","left":"max(1.25rem, 5vw)"}}}} -->
-<div class="wp-block-group" style="padding-right:max(1.25rem, 5vw);padding-left:max(1.25rem, 5vw)"><!-- wp:query {"queryId":1,"query":{"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":"","perPage":12},"tagName":"main","displayLayout":{"type":"flex","columns":3},"layout":{"inherit":true}} -->
-<main class="wp-block-query"><!-- wp:post-template {"align":"wide"} -->
-<!-- wp:post-featured-image {"isLink":true,"width":"100%","height":"318px"} /-->
+<!-- wp:group {"style":{"spacing":{"padding":{"right":"var(--wp--custom--spacing--small, 1.25rem)","left":"var(--wp--custom--spacing--small, 1.25rem)"}}}} -->
+<div class="wp-block-group" style="padding-right:var(--wp--custom--spacing--small, 1.25rem);padding-left:var(--wp--custom--spacing--small, 1.25rem)">
 
-<!-- wp:post-title {"isLink":true,"fontSize":"large"} /-->
+<!-- wp:group {"align":"wide","layout":{"inherit":true}} -->
```
scruffian: Does this block need to be align: wide?

jffng: No, I added a commit to remove some unnecessary alignment declarations.

theme.json
```diff
@@ -1,5 +1,36 @@
 {
 	"version": 1,
+	"customTemplates": [
+		{
+			"name": "blank",
+			"title": "Blank",
+			"postTypes": [
```
jffng, line 7: Just noting that the post types do not work yet, but they will soon: https://github.com/WordPress/gutenberg/pull/35802

jffng (approved):
I made a couple minor adjustments to the index template, I think this is looking good to bring in. Thanks @kjellr !

block-templates/index.html
kjellr: Thank you both!

PR #137, opened 2021-10-20
Add templateParts to theme.json

### LINKED ISSUE #122: Add header and footer template parts to theme.json
When you select a template part in the site editor, and then choose "Replace", you have the option to place the header parts in the footer and vice versa.
Is this intentional?

If it is not intentional, assign the template part to the respective template area in theme.json.

Example:
```
	"templateParts": [
		{
			"name": "header",
			"area": "header"
		},
		{
			"name": "header-large-dark",
			"area": "header"
		}
]
```

carolinan:
**Description**
_Describe the purpose or reason for the pull request_
Add templateParts to theme.json.
Closes #122

**Testing Instructions** 
_Provide steps for testing_
1. In the site editor try typing /footer -the footer should be selectable.
2. try typing /header -the two headers should be selectable
3. Select the current header template part.
4. Select the Replace option in the block toolbar. -The header template that is not in use should be selectable.


jffng (commented):
> In the site editor try typing /footer -the footer should be selectable.

✅
 
> try typing /header -the two headers should be selectable

✅ 

> Select the current header template part.
> Select the Replace option in the block toolbar. -The header template that is not in use should be selectable.

I was able to Replace the header template in the site editor. When I Save, the changes do not persist in the site editor or front-end. Running GB trunk, do you get the same behavior?

carolinan:
Yes, that bug is one of the reasons why I wanted to discuss patterns vs parts.
I have placed the issue in the 5.9  must haves project but I can't say if it will be fixed.
https://github.com/WordPress/gutenberg/issues/35405

jffng:
I see, thanks for pointing me to that issue. 

The "Replace" flow being broken makes me hesitant to merge this. But on the other hand, this does enable adding a new block -> Header / Footer -> "Choose Existing" flow to work correctly (so that footers do not appear when you've selected a header).

theme.json
```diff
@@ -257,5 +257,19 @@
...
 		}
-	}
+	},
+	"templateParts": [
+		{
+			"name": "header",
+			"area": "header"
+		},
+		{
+			"name": "header-large-dark",
```
jffng, line 312: Since these appear in the UI, could we name / capitalize them accordingly? 

```suggestion
			"name": "Header — Dark",
```

or something like that.

carolinan: The `name` is the slug for the file.
I did not include the `Title` because it is not available until Gutenberg 11.8. I was not sure which version is relied on for development.
https://github.com/WordPress/gutenberg/pull/35626

kjellr: We know that Gutenberg 11.9 will be included in 5.9, so anything merged into Gutenberg between now and then is ok to include here today. I think it makes sense to add the title in. Once that's all set, this PR should be ready to go. 👍

carolinan: And what should the title be for the other two items? Just Header and Footer?

carolinan: Anyway I think you have access to editing the PR.

kjellr (approved):
Those names look great. Thanks @carolinan!

PR #141, opened 2021-10-21
Fix index template markup

jffng:
**Description**

There was a bug in the block markup from #136, I missed a closing div on the index template. This PR adds it back. 

I did test with removing this group block entirely, and it seemed fine, but I think it's better to keep the group to contain all the related elements of the post template.

**Testing Instructions** 

Go to the site editor and front-end, verify there are no errors.


kjellr (approved):
👍

PR #145, opened 2021-10-21
Fix stylelint action

jffng:
**Description**

Stylelint 14 was released today and it broke the action. This PR removes it for now, which is I think is fine since there is no CSS in the theme yet.


kjellr (approved):
This seems fine for now, to get the test working again. We should open a followup issue reminder to get it fixed in case we add CSS later though.

Issue #90, opened 2021-10-12
Query Patterns: Followup Fixes


kjellr:
Following up from #84, there are a few query patterns that require followup: 

## Default Posts

- There's no current way to enforce a responsive square shape for featured images. (https://github.com/WordPress/gutenberg/issues/35524) _Edit: Not coming for 5.9._ 

## Simple Blog Posts

- [x] The margin set on the post titles does not appear correct on the front-end. (https://github.com/WordPress/gutenberg/issues/35522)

## Image grid posts

- [x] Post date is supposed to be stacked. This would be solved by the addition of a flex-wrap option. (https://github.com/WordPress/gutenberg/issues/35525)
- [x] Post date can not be formatted into the "11.11.2021" format as originally designed. (https://github.com/WordPress/gutenberg/issues/34762)

## Large post titles

- [x] Post date can not be formatted into the "Nov 11" format as originally designed. (https://github.com/WordPress/gutenberg/issues/34762)

## Irregular posts grid

- [ ] Since this requires use of multiple query blocks, it cannot include post pagination.

carolinan:
For the post dates, this pull request could use more reviews https://github.com/WordPress/gutenberg/pull/35864

carolinan:
@kjellr I really like the irregular posts design. I think it can be kept without pagination. If that's not an option, it can't be done without one query and custom CSS.

On the front the sticky posts are repeated in every column. 
Pattern added to a **post**, front view:
![image](https://user-images.githubusercontent.com/7422055/140073305-6877b348-a435-43b9-b559-5120422efff2.png)

here I replaced the query on **index** with the pattern:
![image](https://user-images.githubusercontent.com/7422055/140074400-a4ff7848-b506-498a-ab5b-4ce3875d0821.png)

I messed up the width here. Chances are other users will do the same. Not sure how users can be guided.

Gutenberg issue for sticky posts: https://github.com/WordPress/gutenberg/issues/35322#issuecomment-946574994

kjellr:
>  There's no current way to enforce a responsive square shape for featured images. ([Post Featured Image]: Allow cropping and/or constraining to a certain ratio gutenberg#35524)

This one doesn't look likely for 5.9 since it would be a new feature and it's past the feature freeze. Going to cross it out for now. 

> Since this requires use of multiple query blocks, it cannot include post pagination.

I'm a little torn on this one. I think the two options are: 

- Do nothing.
[...]

PR #146, opened 2021-10-21
Adjust date formats in query patterns

kjellr:
Addresses two of the checkboxes in https://github.com/WordPress/twentytwentytwo/issues/90:

- Image grid posts: Post date can not be formatted into the "11.11.2021" format as originally designed.
- Large post titles: Post date can not be formatted into the "Nov 11" format as originally designed.

Note that this uses custom (valid) markup in the block to get this to work. When you have one of these selected, the date format dropdown will show an incorrect value. There's an upstream bug filed for that here: https://github.com/WordPress/gutenberg/issues/35854

## Screenshots

Before|After
---|---
Screen Shot 2021-10-21 at 3 21 08 PM|Screen Shot 2021-10-21 at 3 17 45 PM

Before|After
---|---
Screen Shot 2021-10-21 at 3 20 45 PM|Screen Shot 2021-10-21 at 3 20 19 PM

### DIFF
--- a/inc/patterns/query-image-grid.php
+++ b/inc/patterns/query-image-grid.php
@@ -13,7 +13,7 @@ return array(
 					<!-- wp:group {"layout":{"type":"flex","justifyContent":"space-between"}} -->
 					<div class="wp-block-group"><!-- wp:post-title {"isLink":true,"style":{"typography":{"fontFamily":"var:preset|font-family|system-font","fontStyle":"normal","fontWeight":"400"}},"fontSize":"small"} /-->
 
-					<!-- wp:post-date {"format":"F j, Y","style":{"typography":{"fontStyle":"italic","fontWeight":"400"}},"fontSize":"small"} /--></div>
+					<!-- wp:post-date {"format":"m.d.y","style":{"typography":{"fontStyle":"italic","fontWeight":"400"}},"fontSize":"small"} /--></div>
 					<!-- /wp:group -->
 					<!-- /wp:post-template -->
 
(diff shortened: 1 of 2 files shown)


inc/patterns/query-image-grid.php
```diff
@@ -13,7 +13,7 @@
 					<!-- wp:group {"layout":{"type":"flex","justifyContent":"space-between"}} -->
 					<div class="wp-block-group"><!-- wp:post-title {"isLink":true,"style":{"typography":{"fontFamily":"var:preset|font-family|system-font","fontStyle":"normal","fontWeight":"400"}},"fontSize":"small"} /-->
 
-					<!-- wp:post-date {"format":"F j, Y","style":{"typography":{"fontStyle":"italic","fontWeight":"400"}},"fontSize":"small"} /--></div>
+					<!-- wp:post-date {"format":"m.d.y","style":{"typography":{"fontStyle":"italic","fontWeight":"400"}},"fontSize":"small"} /--></div>
```
jffng, line 16: When there are no featured images, this one looks a little off. But okay to address that in a follow-up.

Default alignment:
Screen Shot 2021-10-21 at 5 10 14 PM

Wide alignment:
Screen Shot 2021-10-21 at 5 11 49 PM

kjellr: Ugh yes, this is an issue the grid view for the query block in general. In any case, this is a non-default layout, and it's  called "Image grid posts" so maybe folks won't choose it if it looks bad? 😅

jffng (approved):

PR #154, opened 2021-10-23
Improving error message

Riyadh1734:
Improving a little bit changes to the 404 error page. I  think this is most important to users.


jeffikus:
Hi @Riyadh1734 - I don't think this is good idea to add into the theme, in particular using mention of the pandemic quarantine. I understand the intention of your submission, but this has the potential to bring back some painful memories/experiences for some folks, and I think the "This page could not be found." is sufficient.

block-templates/404.html
```diff
@@ -9,7 +9,7 @@ <h2 class="has-text-align-center" style="font-size:clamp(4rem, 40vw, 20rem);font
 <!-- wp:paragraph {"align":"center"} -->
 <p class="has-text-align-center">This page could not be found.</p>
 <!-- /wp:paragraph -->
-
+<div class="msg">Maybe this page moved? Got deleted? Is hiding out in quarantine? Never existed in the first place?<p>Let's go <a href="#">home</a> and try from there.</p></div>
```
pattonwebz, line 12:
```suggestion
<div class="msg">
    <p>Maybe this page moved, got deleted or never existed in the first place?</p>
    <p>Let's go <a href="/">home</a> and try from there.</p>
</div>
```

I think this should be done more like this - with 2 paragraphs and a real link to home.

jeffikus:
I think, if anything it could be more aligned with prior default themes;
- https://github.com/WordPress/WordPress/blob/master/wp-content/themes/twentytwentyone/404.php
- https://github.com/WordPress/theme-experiments/blob/master/tt1-blocks/block-templates/404.html

There is no need imo for a home link as headers typically have that.

kjellr:
Thanks for the discussion everyone. I agree with @jeffikus on this one. If you'd like to update the default message to be: 

> It looks like nothing was found at this location. Maybe try a search?

That works for me. 👍 

As it is currently, the markup in this PR is not valid block markup though. We'd want to just replace [the string on L10](https://github.com/WordPress/twentytwentytwo/blob/f11b36e8575c1c05606ba3e5e240575b59921a7c/block-templates/404.html#L10).

Riyadh1734:
There is no problem that the home link can be omitted. But there are two lines very important for new users.

block-templates/404.html
Riyadh1734: It can be changed anytime. Thank you for your suggestion.

Riyadh1734:
Hi @jeffikus I agree with you about mentioning pandemic quotes but another two lines are very important for new users & it seems too different but it's helpful.
And also we use "This page could not be found" as sufficient for all. but this time I think "Maybe this page moved? Got deleted? Never existed in the first place? Let's go home and try from there" this message helps new users.

jeffikus:
@Riyadh1734 I hear what you are trying to achieve, my counter point would be then do we list all the possible reasons why a page could not be found? From my perspective that's what the search field is for, the only addition I would consider here is the second part of [this line](https://github.com/WordPress/theme-experiments/blob/master/tt1-blocks/block-templates/404.html#L18), the "Maybe try a search?" part.

pattonwebz:
Just chiming in here to say I am indifferent to the actual message used. My concerns were only to not mention quarantine and that the text is contained within the right (semantically) markup.

Riyadh1734:
@jeffikus lf this is good for all folks.then that's also ok for me. Go for it :)

jffng:
Closing since the message was changed in #182: https://github.com/WordPress/twentytwentytwo/pull/182/commits/66c8adef89d4a58dbee046554a016667c12719f2

Thanks everyone for the improvement.

PR #156, opened 2021-10-25
Use the outer padding to apply the canvas padding and avoid issues with full width blocks

youknowriad:
More details on upcoming Gutenberg PR.


kjellr:
I'm going to close this one for now, since it's outdated and the corresponding PR isn't going to be a part of 5.9.

Issue #150, opened 2021-10-22
Consider overall size of theme directory


Sandstromer:
I noticed the latest trunk, the folder size of the theme is now 13MB unzipped, and even zipped this only reduces to 12.7MB.

Some of the assets bundled with the recently added patterns are massive, e.g. the `birds.mp4` is 3.4MB and `ducks.jpg` is 4.72MB

Is there anything that can be done to reduce some of these further?

#132
#139

kjellr:
This is something we can definitely improve. Regardless of whether we end up using WebP, it sounds like that `ducks.jpg` image is not properly compressed at the moment. 

I'm not sure about feasibility for the initial release, but as proposed in https://github.com/WordPress/twentytwentytwo/issues/12 and https://github.com/WordPress/gutenberg/issues/35364 I do hope that eventually none of those image/video assets end up being bundled in the theme zip.

LittleBigThing:
I compressed the rest of the images.
I have not changed the existing file formats, but

- swallow.jpg would be somewhat smaller if it would be a jpeg.
- we could use (some) WebP images as suggested in #132 

but #159 means already a significant reduction in image sizes.

kjellr:
> I noticed the latest trunk, the folder size of the theme is now 13MB unzipped, and even zipped this only reduces to 12.7MB.

Just wanted to provide some updated numbers after the recent compression PRs. The theme is now 5.3mb unzipped, and 5.2mb zipped.

kjellr:
As per the comment above, I'm going to close this one for now. If anyone has further thoughts to share on this topic, feel free to open up [a new issue on Trac](https://core.trac.wordpress.org/newticket?component=Bundled+Theme). Thanks!

PR #157, opened 2021-10-25
Reduce image size duck

LittleBigThing:
Just a smaller sized duck.jpg. Handy anyway and certainly if the images will get bundled with the theme.

I’d like to note that the other images could be optimised (lossless) further as well. Happy to do it if needed.

Partially solves #150


kjellr (approved):
Thanks, @LittleBigThing! Feel free to take a pass at updating the other images too if you'd like.

Issue #140, opened 2021-10-20
Add duotone presets


kjellr:
We should add duotone presets that use the theme colors. 

- Foreground on Background
- Foreground on Secondary
- Foreground on Tertiary
- Primary on Background
- Primary on Secondary
- Primary on Tertiary

PR #158, opened 2021-10-25
add duotone preset

LittleBigThing:
Add duotone preset to theme.json based on theme colors

Solves #140

I’d like to note that the difference between the foreground and tertiary colors is visually hardly noticeable:

duotone-presets

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -33,6 +33,38 @@
 	],
 	"settings": {
 		"color": {
+			"duotone": [
+                {
+                    "colors": [ "#000000", "#ffffff" ],
+                    "slug": "foreground-and-background",
+                    "name": "Foreground and background"
+                },
+				{
+                    "colors": [ "#000000", "#ffe2c7" ],
+                    "slug": "foreground-and-secondary",
+                    "name": "Foreground and secondary"
+                },
+				{
+                    "colors": [ "#000000", "#f6f6f6" ],
+                    "slug": "foreground-and-tertiary",
+                    "name": "Foreground and tertiary"
+                },
+				{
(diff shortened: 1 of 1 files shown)


kjellr:
Thanks for the PR, @LittleBigThing! 

> I’d like to note that the difference between the foreground and tertiary colors is visually hardly noticeable:

That's ok by me — it's subtle, but there will be use cases when one is better than the other (for example, if the image is on a tertiary-colored background, you'll want to use the tertiary-colored duotone.

theme.json
```diff
@@ -33,6 +33,38 @@
 	],
 	"settings": {
 		"color": {
+			"duotone": [
+                {
+                    "colors": [ "#000000", "#ffffff" ],
```
kjellr, line 38: Should these be defined using the CSS variables instead? Like we do for gradients? 

https://github.com/WordPress/twentytwentytwo/blob/83badf191b8922f9d825076cf49deef1b64a8c67/theme.json#L39

Or does that not work?

LittleBigThing: No, it didn’t seem to work for me like this:

```
"duotone": [
                {
                    "colors": [ "var(--wp--preset--color--foreground)", "var(--wp--preset--color--background)" ],
                    "slug": "foreground-and-background",
                    "name": "Foreground and background"
                }
```

Maybe I'm doing something wrong?
The palette is defined in hex codes as well.

kjellr: Yeah, it looks like it's not possible at this point. 👍

kjellr (approved):
Looks good, thanks @LittleBigThing!

PR #159, opened 2021-10-25
Compress images

LittleBigThing:
The existing images in the assets/patterns folder are compressed (saving a lot of data!), keeping the same image format.

Partially solves #150


kjellr:
These look good, thanks @LittleBigThing! I made a couple small edits, and will merge in from here!

- I made `swallow.png` into a `jpg` as you suggested.
- I generated a new version of `hummingbird-flying.jpg`, since I noticed there was a visual bug in the bottom of the existing one. This one's slightly bigger than the one you'd included here, since I realized it was looking fuzzy on larger screens in the context of the pattern.

PR #160, opened 2021-10-25
compress birds.mp4

LittleBigThing:
Reduce the size of birds.mp4 without significant reduction of visual quality (IMO).

#150


kjellr:
This seems acceptable, thank you!

PR #163, opened 2021-10-26
Add simple version of button hover state

### LINKED ISSUE #99: Buttons need hover, active, and focus states

jffng:
**Description**

This PR adds a default button hover state that just changes the background color. Works for both default and outline style buttons. Falls back to no hover state if a color customizations have been made.

Alternative to #162, which is more complicated and will probably not work with the hover / active state capabilities coming later in Gutenberg. Solves #99.

**Screenshots**

https://user-images.githubusercontent.com/5375500/138921352-ea5907d4-20d6-4d7f-80d3-e29b025d4249.mp4

**Testing Instructions** 

1. Test default and outline style buttons in the editor and front-end
2. Adjust the padding on a global and block level, verify the hover states still work
3. Adjust the colors on a global and block level, verify the hover state does not work

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -13,13 +13,10 @@ if ( ! function_exists( 'twentytwentytwo_styles' ) ) :
 	 * Enqueue scripts and styles.
 	 */
 	function twentytwentytwo_styles() {
-		// Register theme stylesheet.
-		wp_register_style( 'twentytwentytwo-style', '' );
+		// Register theme stylesheet
+		wp_enqueue_style( 'twentytwentytwo-style', get_stylesheet_uri() );
 		// Add styles inline.
 		wp_add_inline_style( 'twentytwentytwo-style', twentytwentytwo_get_font_face_styles() );
-		// Enqueue theme stylesheet.
-		wp_style_add_data( 'twentytwentytwo-style', 'path', get_template_directory() . '/style.css' );
-		wp_enqueue_style( 'twentytwentytwo-style' );
 	}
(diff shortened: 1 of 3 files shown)


functions.php
```diff
@@ -13,13 +13,10 @@ function twentytwentytwo_support() {
 	 * Enqueue scripts and styles.
 	 */
 	function twentytwentytwo_styles() {
-		// Register theme stylesheet.
-		wp_register_style( 'twentytwentytwo-style', '' );
+		// Register theme stylesheet
+		wp_enqueue_style( 'twentytwentytwo-style', get_stylesheet_uri() );
```
jffng, line 17: Can update this to use the inline loading method if we're happy with the styles.

jffng:
A couple things to note even with this "simplified" approach:

- The only way to change the hover color is to edit the custom variable in theme.json OR change the background + foreground colors in the global palette
- If the user changes their button color globally, the hover state will still apply

Are these tradeoffs okay? What do you think is the path forward here, given that we will not likely get hover state control in 5.9 @kjellr?

kjellr:
> The only way to change the hover color is to edit the custom variable in theme.json OR change the background + foreground colors in the global palette
> If the user changes their button color globally, the hover state will still apply

Yeah, both of those don't seem great to me. I do expect that we'll have a UI control for picking the hover color eventually, and that seems like it will cause issues with this PR. So I think I lean towards the more complicated solution here. 😕

kjellr:
Closing in favor of #275

Issue #151, opened 2021-10-22
Determine strategy for templates that use sidebars


kjellr:
A couple of the theme mockups feature a simple sidebar on the left or right: 

Left Sidebar|Right Sidebar
---|---
<img width="1200" alt="Blog" src="https://user-images.githubusercontent.com/1202812/138500019-d25ad7a7-2633-4e06-8661-02cbb54af72e.png">|<img width="1200" alt="Blog-1" src="https://user-images.githubusercontent.com/1202812/138500017-280597a9-fd03-4878-8558-a0e2888ddff1.png">

I think it's worthwhile to making these templates available, but I'm not certain the best way to implement them within the boundaries of what will launch with 5.9. Ideally, users would be able to switch out their entire `index.html`, `single.html`, and `page.html` templates for versions of these. But FSE doesn't currently include the concept of swapping out entire templates except on a single post/page basis. 

I _think_ this seems like the best route for now:

- Offer a few alternate page templates that map to these mockups. For example:
    -  "Left sidebar: Post list"
    -  "Left sidebar: Single post"
    -  "Left sidebar: Single page"
- We'd also bundle a "sidebar" template part so that would stay in sync across these templates. 
- Users could then assign one of those templates to static pages as needed. 

It's not totally ideal, but I think it would technically work? The other option is basically to just build these as patterns and folks can paste them into the post/page content using our "Blank" page template. 

I'm curious if other folks have ideas!

overclokk:
Personally what I think is less it's better, because there are more combination with content and sidebar, here a few;

```
sidebar-sidebar-content
sidebar-content-sidebar
content-sidebar-sidebar

sidebar-content
content-sidebar

content
```

I'm experimenting a theme with only `singular.html` for page and post with only the `content` and adding a pattern with `content-sidebar`, so you can add the template you need only if you need it, switching sidebar left to right is a matter of a click on the arrow button.

carolinan:
> I think it would technically work? The other option is basically to just build these as patterns and folks can paste them into the post/page content using our "Blank" page template.
> 
> I'm curious if other folks have ideas!

Yes using template parts for it works.
A pattern also works well when users do not want to use the same sidebar content.

PR #164, opened 2021-10-26
Add sidebar patterns

kjellr:
This is more or less an MVP of #151. It just adds these as simple "Page" patterns, designed to be used with our "Blank" page template. I think it might make sense to just do this until there's a clearer solution in Gutenberg.

Posts with Left Sidebar|Posts with Right Sidebar
---|---
Screen Shot 2021-10-26 at 2 14 40 PM|Screen Shot 2021-10-26 at 2 17 49 PM


kjellr:
You may notice an alignment issue with the site title issue while testing. I've submitted a fix upstream: https://github.com/WordPress/gutenberg/pull/35971

inc/patterns/page-sidebar-left.php
```diff
@@ -0,0 +1,77 @@
...
+
+					<!-- wp:spacer {"height":16} -->
+					<div style="height:16px" aria-hidden="true" class="wp-block-spacer"></div>
+					<!-- /wp:spacer -->
+
+					<!-- wp:site-logo {"width":60} /--></div>
+					<!-- /wp:column -->
+
+					<!-- wp:column {"width":"70%"} -->
+					<div class="wp-block-column" style="flex-basis:70%"><!-- wp:query {"queryId":31,"query":{"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","sticky":"","inherit":false,"perPage":12},"displayLayout":{"type":"flex","columns":3},"layout":{"inherit":true}} -->
```
jffng: Nit, I don't think the `queryId` is necessary?

inc/patterns/page-sidebar-right.php
```diff
@@ -0,0 +1,79 @@
+<?php
+/**
+ * Posts with right sidebar block pattern
```
jffng: This one is weird on mobile because the sidebar stacks below the query, we'd need to supply a `flex-direction: row-reverse;` on smaller devices or something...

inc/patterns/page-sidebar-right.php
```diff
@@ -0,0 +1,79 @@
...
+	'categories' => array( 'twentytwentytwo-pages' ),
+	'content'    => '<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"max(1.25rem, 5vw)","right":"max(1.25rem, 5vw)","bottom":"max(1.25rem, 5vw)","left":"max(1.25rem, 5vw)"}}},"layout":{"inherit":true}} -->
+					<div class="wp-block-group alignfull" style="padding-top:max(1.25rem, 5vw);padding-right:max(1.25rem, 5vw);padding-bottom:max(1.25rem, 5vw);padding-left:max(1.25rem, 5vw)"><!-- wp:columns {"align":"wide","style":{"spacing":{"margin":{"top":"0px","bottom":"0px"},"blockGap":"5%"},"elements":{"link":{"color":{"text":"var:preset|color|primary"}}}},"textColor":"primary"} -->
+					<div class="wp-block-columns alignwide has-primary-color has-text-color has-link-color" style="margin-top:0px;margin-bottom:0px"><!-- wp:column {"width":"66.66%"} -->
+					<div class="wp-block-column" style="flex-basis:66.66%"><!-- wp:query {"queryId":37,"query":{"perPage":"5","pages":0,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":false},"layout":{"inherit":true}} -->
```
jffng: Same comment about the `queryId`

jffng (commented):
> designed to be used with our "Blank" page template. I think it might make sense to just do this until there's a clearer solution in Gutenberg.

👍 inserting them in this way is pretty straightforward.

The alternative is making them entire page templates, which feels a bit heavy handed, especially since AFAIK there's no way in the site editor to assign an existing template to a template location (e.g. I want all pages to use the `page-sidebar.html` template)

inc/patterns/page-sidebar-right.php
kjellr: Yeah. 😕 @jasmussen, do you have any ideas for this one? It uses a standard columns block, and we'd like it to stack the second column on top when it wraps for small screens. 

If we can't achieve this now, I can always just switch this to use a left column instead.

inc/patterns/page-sidebar-right.php
kjellr: Thanks, this has been removed.

inc/patterns/page-sidebar-left.php
kjellr: Thanks, this has been removed.

inc/patterns/page-sidebar-right.php
jasmussen: I briefly pondered if we could use the Row block for some automatic wrapping here, but it feels too sensitive at this point. And so outside of writing CSS to help accomplish this, I can't think of a good solution. row-reverse is also risky to use, as I don't think it changes the tab order from matching that of the DOM, which means the flow can become a bit confusing just when tabbing through.

kjellr: Ok cool, I'll switch this template to be a second left-sidebar template in the meantime. I still think it'll be useful!

kjellr:
Ok, this is ready for another round of testing: 

Grid of posts with left sidebar|Blog posts with left sidebar
---|---
Screen Shot 2021-11-01 at 12 19 23 PM|Screen Shot 2021-11-01 at 12 20 22 PM

jffng (approved):
Looking great.

syllie:
Just checking in - there will not be a sidebar-right due to encountered issues?

kjellr:
Not for these two patterns at least. 

We may still create more patterns that include a right sidebar, and users are still free to create sites with a right sidebar manually though.

syllie:
Thanks for your reply @kjellr . I was hoping tt2 would come up with a solution for real world cases (where there are still sidebars in use). I do understand the complications. Thanks for the update.

kjellr:
To be clear, this isn't saying right sidebars will not be possible with the theme. Just that these two specific patterns will now use a left sidebar.

PR #165, opened 2021-10-26
Cleanup the assets directory

kjellr:
This PR does a little cleanup in our assets directory: 

- It removes the extraneous `fonts/source-serif-pro` directory, instead moving the fonts directly into the `fonts` directory. We only plan on bundling one font with the theme, so there's no need to have this be two levels deep. 
- It renames `img` to `images`, which seems more in sync with the naming of our other directories.
- It moves the `birds.mp4` video from the `images` directory into a new `videos` directory. 
- It renames all our images so the names make a little more sense.


jffng (approved):
Thank you!!

PR #124, opened 2021-10-16
Footer patterns: allow translators to control the whole credit string

audrasjb:
**Description**

For footer patterns with WordPress credit links, translators will need to be able to control the whole sentence. 

Here is the current implementation:
```
esc_html__( 'Proudly powered by ', 'twentytwentytwo' ) . '<a href="' . esc_url( __( 'https://wordpress.org', 'twentytwentytwo' ) ) . '" rel="nofollow">WordPress</a></p>
```

Depending on their language, translator may need to move the "WordPress" link at the beginning of the sentence, for example. Therefore, we need to use a placeholder for the link.

Proposal:
```
sprintf(
	/* Translators: WordPress link. */
	esc_html__( 'Proudly powered by %s', 'twentytwentytwo' ),
	'<a href="' . esc_url( __( 'https://wordpress.org', 'twentytwentytwo' ) ) . '" rel="nofollow">WordPress</a>'
)
```

I tested this change with the related patterns and and it doesn't change anything to the current rendering.

PR #167, opened 2021-10-27
I18n: Allow translators to control the whole credit string in footer patterns

dimadin:
In #124, few strings were changed so that they are no longer broken and instead use placeholder. This PR covers strings that were missed in #124.


kjellr:
Thanks for catching these, @dimadin!

kjellr (approved):

PR #168, opened 2021-10-27
Try a really simple button hover state

kjellr:
Alternative to #162, #163. 
Solves #99.

I don't really like this treatment, but it's not horrible. It's also just a couple CSS rules and it doesn't break anything. 🤷‍♂️

buttons


jffng (commented):
I think this is mostly a design question. 

From a developer perspective, it's true it will not break since we aren't changing the colors. But if I wanted to change the hover state I'd still need to edit the CSS. At that point, it seems like a nicer experience and result to just change two values in theme.json (as would be the case in #163).

kjellr:
Yeah... I don't like it. It was worth exploring, but I'll close for now.

Issue #116, opened 2021-10-15
Add starter content


kjellr:
On a fresh install, the theme should generally end up matching [the demo site](https://2022.wordpress.net) by default: 

![Screen Shot 2021-10-15 at 10 01 26](https://user-images.githubusercontent.com/1202812/137500559-96f76e6e-00b1-41c5-86cd-a858f36e3605.png)

- Add a static homepage, with the `home` slug (so that it inherits our `page-home.html` template. The content of this page should match the demo above.
- Maybe a few posts to match the ones on the demo site.

In previous themes, we created starter content to help set folks up with this. Is this the best way to handle this for block themes?

kjellr:
I opened a Gutenberg issue to explore this there as well: https://github.com/WordPress/gutenberg/issues/35680

carolinan:
It was actually not so easy to get starter content to work even with the current "live preview" button before the theme is activated.

-I ended up using `page-slug.html` templates instead of the "home/frontpage/blog/"  because there is no conditional like in PHP where we can show single page content if it's set to show a page, and query when it is set to show posts.

Using a query and setting it to `inherit `is one way but it did not work for me since I wanted to show the full content for the page, and only show excerpt for the post list...

mburridge:
Is it normal to have starter content? I don't recall ever installing a theme that added content to a site. Do you intend that the starter content will be "hard coded" into the .html files to be over-ridden by templates created in the editor? Personally I don't think I'd want that 'extra job', I want a theme to be blank so that I can just start adding content.

carolinan:
Starter content is:

- Visible in the preview in the theme directory of WordPress.org, without it, you can only show the blog, not a custom homepage.
- Only available for users on fresh, empty WordPress installations.
- Optional, you can preview it and then choose to publish it.

mburridge:
Thanks for that @carolinan. I've never used starter content before when I've built sites, but I've been[ digging in](https://make.wordpress.org/core/2016/11/30/starter-content-for-themes-in-4-7/) and learning more.

It seems to require the customiser. How is this going to work for an FSE theme?

And from a development point of view, would starter content still be hooked to `add_theme_support( 'starter-content', ....)`, or does it need to be implemented differently in an FSE theme?

carolinan:
> It seems to require the customiser. How is this going to work for an FSE theme?

It does, but it is only the menu item that is removed. The starter content can still be previewed and activated via the themes screen, using the Live Preview button.

[...]

Issue #92, opened 2021-10-12
Include the flying bird image in the homepage header


kjellr:
The original mockup for the theme includes this flying bird image on the homepage template: 

<img width="622" alt="Screen Shot 2021-10-12 at 1 33 06 PM" src="https://user-images.githubusercontent.com/1202812/137002726-9d8d29ed-6fed-4a2e-994e-108cbbcecfa4.png">

This has been left out for now while we figure out the best way to include it. Here's the situation: 

Normally, we'd bundle this image in the theme files and refer to it in the PHP templates. This isn't possible in block templates, since the image block can't take a dynamic URL.  It will work fine in patterns though, where we can use PHP to find the right path. So a workaround like #56 may be possible.

We _could_ put it up on the WP.org CDN, or in the Pattern Directory's media library, and hotlink to it. But it feels non-ideal to have to require that 3rd-party request by default on the front-end. 

I think there's also something to be said about _not_ including this image here by default, since approx 99% of folks will not actually keep this bird image on the front page of their website. 😂

I'm opening this issue for discussion and to find a solution one way or another.

carolinan:
Upstream issue: https://github.com/WordPress/gutenberg/issues/31815

MaggieCabrera:
Well, if the pattern block actually ends up making it, that would solve it, right? A pattern block ends up being the solution to so many things!

kjellr:
Yeah, the pattern block would solve it, but it seems like a weird workaround to me. 🤔

kjellr:
This should now be possible by using the Pattern block. I'm changing this issue's title + tags as a result.

PR #169, opened 2021-10-28
Add basic starter content

kjellr:
⚠️ Do not merge!

---

We currently don't have a way to get starter content to work for block themes. But here's a quick PR that implements some very basic starter content, the old way. I figured I'd create this in case it comes in handy for testing.

So far, all it does is set up  a static homepage and  blog page, and then provide a couple patterns on the homepage to better match [the demo site](https://2022.wordpress.net/). The homepage will look way nicer once we get https://github.com/WordPress/twentytwentytwo/issues/92 figured out.

## To test: 

- Start a fresh site.
- Install the theme, don't touch any other settings. 
- Manually visit the customizer. (`{site-url]/https://spurred-storm.jurassic.ninja/wp-admin/customize.php`)
- Press the Publish button.
- Starter content should be applied. 

## Screenshot: 

spurred-storm jurassic ninja_

Related: https://github.com/WordPress/twentytwentytwo/issues/116

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -5,6 +5,17 @@ if ( ! function_exists( 'twentytwentytwo_support' ) ) :
 		// Adding support for core block visual styles.
 		add_theme_support( 'wp-block-styles' );
 	}
+
+	/*
+	* Adds starter content to highlight the theme on fresh sites.
+	* This is done conditionally to avoid loading the starter content on every
+	* page load, as it is a one-off operation only needed once in the customizer.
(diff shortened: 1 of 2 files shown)


carolinan:
The live preview on the theme installation page still leads to the customizer. So there is no change compared to classic themes.

kjellr:
Closing this one, since #243 provides a good-enough first experience that should also work in the theme showcase.

PR #173, opened 2021-10-28
Update Dark Header to use Heading instead of Site Tagline

### LINKED ISSUE #170: Reconsider use of Site Tagline for the large header on the homepage
Currently, the large header area on the homepage uses the Site Tagline block: 

Screen Shot 2021-10-28 at 9 03 55 AM

I've been going back and forth about whether this is a good idea, so I thought I'd write down my concerns. the alternative is just to have it be a heading block. 

Pro:
- It's pre-populated with something that's potentially relevant on theme switch.
- If users are confused about how to customize it, there's a second place they can edit it: `Settings > Reading`. (However, once there's [a bird image](https://github.com/WordPress/twentytwentytwo/issues/92) in there, folks will learn to edit the template part in order to remove/change that image anyway.)

Con:
- It's a little unexpected — usually the site tagline is nowhere near this prominent. 
- Users might want their site tagline (as used for SEO and in the title tag) to be different than the big message they use to greet visitors.
- Semantically, it's a paragraph, but visually it seems more like a headline. 
- The site tagline does not support rich text controls, so we can't actually achieve [the italicized part of the original mockup](https://cloudup.com/cQMwENCI3KO).
- Once we have the bird image in there, I think "Just another WordPress site" is going to be a really weird text pair for it. Plus, it wraps oddly. 😕 

I'm interested in other folks' thoughts!

kjellr:
Resolves #170.

This PR: 

- Replaces the "Site Tagline" with a standard Heading block in both the `header-large-dark` template part and pattern. 
- Adds the (compressed!) bird image to the pattern. 
- Reorders the header patterns so that this one appears closer to the top. (Since it's one of the defaults, I think it should be easier to find.)

## Screenshots:
Screen Shot 2021-10-28 at 11 52 13 AM
Screen Shot 2021-10-28 at 11 50 09 AM


kjellr:
Actually, I'm seeing some markup issues with the PR — please hold off on a review for the moment.

kjellr:
Ok, should be ready for review now. 😄

inc/patterns/header-large-dark.php
```diff
@@ -6,19 +6,27 @@
...
 					<!-- /wp:navigation --></div>
 					<!-- /wp:group -->
 
-					<!-- wp:site-tagline {"align":"wide","style":{"typography":{"fontFamily":"var:preset|font-family|source-serif-pro","fontWeight":"300","fontSize":"clamp(4rem, 8vw, 6.25rem)","lineHeight":"1.15"}}} /--></div>
+					<!-- wp:heading {"align":"wide","style":{"typography":{"fontSize":"clamp(4rem, 8vw, 6.25rem)","lineHeight":"1.15"}}} -->
+					<h2 class="alignwide" style="font-size:clamp(4rem, 8vw, 6.25rem);line-height:1.15">' . wp_kses_post( __( '<em>The Hatchery</em>: a blog about my adventures in bird watching', 'twentytwentytwo' ) ) . '</h2>
+					<!-- /wp:heading --></div>
+					<!-- /wp:group -->
+
+					<!-- wp:image {"align":"full","id":55,"sizeSlug":"full","linkDestination":"none"} -->
```
jffng: I think the `id` and respective CSS class can be removed?

jffng (approved):
One small comment, LGTM.

inc/patterns/header-large-dark.php
kjellr: 👍 They can! Removed in https://github.com/WordPress/twentytwentytwo/pull/173/commits/edbf54013be60732e7e65fcbb0425299163b53ad.

PR #182, opened 2021-10-29
Replace template strings with translatable patterns

### LINKED ISSUE #30: String translation in templates

jffng:
**Description**

This PR adds a 404 pattern to the general category, and implements it patterns inside the 404 template. It also replaces the contents of `footer.html` with the `footer-default` pattern. 

This is done so that the strings can be marked for translation.

Closes #30.

**Screenshots**

There should be no visual change.

**Testing Instructions** 

Go to a 404 page and verify the contents appear as expected.


inc/patterns/general-404.php
```diff
@@ -0,0 +1,18 @@
...
+ * 404 content.
+ */
+return array(
+	'title'      => __( '404 content', 'twentytwentytwo' ),
+	'categories' => array( 'twentytwentytwo-general' ),
+	'content'    => '<!-- wp:heading {"style":{"typography":{"fontSize":"clamp(4rem, 40vw, 20rem)","fontWeight":"200","lineHeight":"1","spacing":{"margin":{"top":"0","bottom":"0"}}}},"className":"has-text-align-center"} -->
+					<h2 class="has-text-align-center" style="font-size:clamp(4rem, 40vw, 20rem);font-weight:200;line-height:1;margin-top:0;margin-bottom:0;">404</h2>
+					<!-- /wp:heading -->
+					<!-- wp:paragraph {"align":"center"} -->
+					<p class="has-text-align-center">' . esc_html( 'This page could not be found. Maybe try a search?', 'twentytwentytwo' ) . '</p>
```
jffng: Implemented the suggestion from here https://github.com/WordPress/twentytwentytwo/pull/154#issuecomment-951122568

inc/patterns/general-404.php
```diff
@@ -0,0 +1,18 @@
+<?php
+/**
+ * 404 content.
+ */
+return array(
+	'title'      => __( '404 content', 'twentytwentytwo' ),
+	'categories' => array( 'twentytwentytwo-general' ),
+	'content'    => '<!-- wp:heading {"style":{"typography":{"fontSize":"clamp(4rem, 40vw, 20rem)","fontWeight":"200","lineHeight":"1","spacing":{"margin":{"top":"0","bottom":"0"}}}},"className":"has-text-align-center"} -->
```
jffng: I had to zero the top and bottom margins here because the markup changed slightly when inserting it as a block.

kjellr:
@jffng rather than implement a separate credit line block, can we just make the entirety of the `footer.html` template be a pattern block that points to the `footer-default` pattern?

https://github.com/WordPress/twentytwentytwo/blob/trunk/inc/patterns/footer-default.php

jffng:
I just tried that and it causes a layout issue because of the additional wrapping div introduced by the pattern block:

Screen Shot 2021-10-29 at 2 44 07 PM

jffng:
I have an idea for a fix in Gutenberg which would remove this wrapping div entirely.

kjellr:
> I have an idea for a fix in Gutenberg which would remove this wrapping div entirely.

Cool — ideally these should be entirely seamless replacements.

inc/patterns/general-404.php
```diff
@@ -0,0 +1,18 @@
+<?php
+/**
+ * 404 content.
+ */
+return array(
+	'title'      => __( '404 content', 'twentytwentytwo' ),
+	'categories' => array( 'twentytwentytwo-general' ),
+	'content'    => '<!-- wp:heading {"style":{"typography":{"fontSize":"clamp(4rem, 40vw, 20rem)","fontWeight":"200","lineHeight":"1"}},"className":"has-text-align-center"} -->
+					<h2 class="has-text-align-center" style="font-size:clamp(4rem, 40vw, 20rem);font-weight:200;line-height:1">' . esc_html__( '404', 'twentytwentytwo' ) . '</h2>
```
kjellr: I'm really not sure, but should this have context alongside it?

```suggestion
					<h2 class="has-text-align-center" style="font-size:clamp(4rem, 40vw, 20rem);font-weight:200;line-height:1">' . esc_html( _x( '404', 'Error code for a webpage that is not found.', 'twentytwentytwo' ) ) . '</h2>
```

kjellr:
Assuming https://github.com/WordPress/gutenberg/pull/36090 lands, this PR seems to be in good shape. 👍

inc/patterns/general-404.php
jffng: I think adding context makes sense. What's the difference between using `_x` and this method: https://github.com/WordPress/twentytwentytwo/blob/trunk/inc/patterns/footer-default.php#L15-L17

jffng:
Merging since the above Gutenberg PR landed.

PR #183, opened 2021-10-29
Add the bird image to the homepage header

### LINKED ISSUE #92: Include the flying bird image in the homepage header

kjellr:
Fixes #92. 

This PR uses the new pattern block to place the `header-large-dark` pattern in the homepage's header. Like #182, this PR relies on https://github.com/WordPress/gutenberg/pull/36090 to get it to appear correctly. 

Before|After
---|---
Screen Shot 2021-10-29 at 4 16 35 PM|Screen Shot 2021-10-29 at 4 15 59 PM


jffng (approved):
Nice, love that we can remove repetitive code. Assuming the Gutenberg PR lands, this LGTM.

kjellr:
@jffng just pushed a quick update to re-use the `header.html` template part (as per the method in https://github.com/WordPress/twentytwentytwo/pull/55, which I think we still need to use). Could use another review just to be sure. 👍

inc/patterns/header-large-dark.php
```diff
@@ -8,17 +8,7 @@
...
-					<div class="wp-block-group"><!-- wp:site-logo {"width":64} /-->
-
-					<!-- wp:site-title {"style":{"typography":{"fontStyle":"italic","fontWeight":"400"}}} /--></div>
-					<!-- /wp:group -->
-
-					<!-- wp:navigation {"itemsJustification":"right","isResponsive":true} -->
-					<!-- wp:page-list /-->
-					<!-- /wp:navigation --></div>
-					<!-- /wp:group -->
+					<div class="wp-block-group alignfull" style="padding-top:0px;padding-right:max(1.25rem, 5vw);padding-bottom:0px;padding-left:max(1.25rem, 5vw)"><!-- wp:template-part {"slug":"header","tagName":"header","theme":"twentytwentytwo","layout":{"inherit":true},"align":"wide"} /-->
```
carolinan: Why is the theme attribute needed?

kjellr: It throws a "template part not found" error otherwise.

carolinan:
To get the dark header, I selected the header template part and used replace.
<del>The image of the bird is showing, but the dark header is not the same width as earlier. Only the last, the inner most group block, is set to full width. 🤔 </del>
Ah, that's why the GB pr was needed.

carolinan:
Also, the preview is not working for the dark header. which needs to be an upstream issue, if one hasn't already been opened

kjellr:
> Also, the preview is not working for the dark header. which needs to be an upstream issue, if one hasn't already been opened

What are you seeing there? It did work for me, but it took a minute for the top of the header to appear: 

Screen Shot 2021-11-01 at 9 30 26 AM

carolinan:
I mean this preview not the pattern preview in the sidebar:
image

carolinan:
Who has time to wait a full minute? :) (waiting doesn't help for this preview...)

carolinan:
The problem with the `theme attribute` is that if I move the theme into a subfolder like wp-content/themes/theme-experiments/, this message is shown:
`Template part has been deleted or is unavailable: header`

inc/patterns/header-large-dark.php
carolinan: Then that is a bug that needs to be fixed, because with it, it shows a template part not found if the theme is in a sub folder -that's why we stopped using the attribute for all template parts.

kjellr: For what it's worth, I moved the theme into a subfolder and things continued to work as expected.

kjellr:
> I mean this preview not the pattern preview in the sidebar:

I'm already seeing errors with this preview in `trunk`, so I'm not sure it should hold up this PR: 

Screen Shot 2021-11-02 at 9 51 36 AM

> The problem with the theme attribute is that if I move the theme into a subfolder like wp-content/themes/theme-experiments/, this message is shown:
Template part has been deleted or is unavailable: header

I moved the theme into a subfolder and things continued to work as expected. 

Even so, I tried out an alternate approach that uses a pattern just for the text + bird image, but it requires more code and also showed preview errors. So I'm not totally sure it's worth it. 

Would someone mind testing this out in a subfolder to see if it works or not? If it's still a bug, I can push that alternate approach for now.

jffng:
This is what I see when I insert the dark header and the theme is a subfolder: 

Screen Shot 2021-11-02 at 1 06 33 PM

kjellr:
Thanks for testing. I wonder why it worked for me. 🤔 

Anyway, I pushed that second, more elaborate method. Would you mind giving that a try next?

jffng:
> Anyway, I pushed that second, more elaborate method. Would you mind giving that a try next?

This worked for me with the theme in a subfolder, no template part not found error.

inc/patterns/hidden-heading-and-bird.php
```diff
@@ -0,0 +1,21 @@
+<?php
+/**
+ * Heading and bird image
+ *
+ * This pattern is used only for translation
+ * and to refernce a dynamic image URL. It does
```
jffng:
```suggestion
 * and to reference a dynamic image URL. It does
```

block-template-parts/header-large-dark.html
```diff
@@ -1,9 +1,7 @@
...
-<h2 class="alignwide" id="the-hatchery-a-blog-about-adventures-in-bird-watching" style="font-size:clamp(4rem, 8vw, 6.25rem);line-height:1.15"><em>The Hatchery</em>: a blog about my adventures in bird watching</h2>
-<!-- /wp:heading --></div>
-<!-- /wp:group --></div>
-<!-- /wp:group -->
\ No newline at end of file
+<div class="wp-block-group alignfull has-background-color has-foreground-background-color has-text-color has-background has-link-color" style="margin-bottom:8rem;padding-top:0px;padding-right:0px;padding-bottom:8rem;padding-left:0px"><!-- wp:group {"align":"full","style":{"spacing":{"padding":{"right":"max(1.25rem, 5vw)","left":"max(1.25rem, 5vw)","top":"0px","bottom":"0px"}}},"layout":{"inherit":true}} -->
+<div class="wp-block-group alignfull" style="padding-top:0px;padding-right:max(1.25rem, 5vw);padding-bottom:0px;padding-left:max(1.25rem, 5vw)"><!-- wp:template-part {"slug":"header","tagName":"header","layout":{"inherit":true},"align":"wide"} /--></div>
+<!-- /wp:group -->
+
+<!-- wp:pattern {"slug":"twentytwentytwo/header-heading-and-bird"} /--></div>
```
jffng:
```suggestion
<!-- wp:pattern {"slug":"twentytwentytwo/hidden-heading-and-bird"} /--></div>
```

jffng (approved):
Two small typo fixes and this LGTM.

Issue #179, opened 2021-10-29
Replace "Post Comments" block with the Comments Query Loop block in templates


kjellr:
Post Comments will remain [marked as "experimental" for 5.9](https://github.com/WordPress/gutenberg/pull/35979#issuecomment-954364262). We should replace with the new "Comments Query Loop" block from https://github.com/WordPress/gutenberg/pull/35231.

scruffian:
I made a start on this here: https://github.com/WordPress/twentytwentytwo/pull/192

jffng:
Closing because Post Comments will be available in 5.9, Comments Query Loop is still in early stages.

PR #192, opened 2021-11-01
Change Post Comments to Query Comments

scruffian:
**Description**

Post Comments isn't shipping in 5.9 so trying Query Comments instead as proposed in https://github.com/WordPress/twentytwentytwo/issues/179

**Screenshots**
Screenshot 2021-11-01 at 17 46 10

**Testing Instructions** 

1. Open a post/page with comments and see what they look like

### DIFF
--- a/block-template-parts/comments.html
+++ b/block-template-parts/comments.html
@@ -1,0 +1,9 @@
+<!-- wp:comments-query-loop -->
+<div class="wp-block-comments-query-loop"><!-- wp:comment-template -->
+	<!-- wp:post-comment-author {"isLink":true} /-->
+	<!-- wp:post-comment-date {"format":"F j, Y g:i a","isLink":true,"fontSize":"small"} /-->
+	<!-- wp:post-comment-content /-->
+
+	<!-- /wp:comment-template -->
+</div>
+<!-- /wp:comments-query-loop -->
--- a/block-templates/page-no-separators.html
+++ b/block-templates/page-no-separators.html
@@ -11,7 +11,7 @@
 
 <!-- wp:group {"layout":{"inherit":true}} -->
 <div class="wp-block-group">
-<!-- wp:post-comments {"style":{"spacing":{"padding":{"top":"var(--wp--custom--spacing--medium, 6rem)"}}}} /--></div>
+<!-- wp:template-part {"slug":"comments","style":{"spacing":{"padding":{"top":"var(--wp--custom--spacing--medium, 6rem)"}}}} /--></div>
 <!-- /wp:group --></main>
 <!-- /wp:group -->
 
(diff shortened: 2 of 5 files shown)


kjellr:
Does this need to have the post comments form in it too? I don't see the ability to add comments otherwise. 

Also... how much styling is possible here? We'd like it to look like it does in `trunk`.

scruffian:
I can't find a way to add the form without using the post comments block, which isn't shipping in 5.9. I'm not sure if this will be possible...

kjellr:
I'm able to place it in manually via the site editor, and that generally seems to work?

```
<!-- wp:comments-query-loop -->
<div class="wp-block-comments-query-loop"><!-- wp:comment-template -->
<!-- wp:post-comment-author {"isLink":true} /-->

<!-- wp:post-comment-date {"format":"F j, Y g:i a","isLink":true,"fontSize":"small"} /-->

<!-- wp:post-comment-content /-->
<!-- /wp:comment-template --></div>
<!-- /wp:comments-query-loop -->

<!-- wp:post-comments-form /-->
```

jffng:
> I'm able to place it in manually via the site editor, and that generally seems to work?

I'm confused about how this is expected to work in 5.9. 

The `post-comment-author`, `post-comment-date`, `post-comment-content`,  and `post-comments-form` are all [currently marked as experimental](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/index.js#L245-L257).

block-template-parts/comments.html
```diff
@@ -0,0 +1,9 @@
+<!-- wp:comments-query-loop -->
+<div class="wp-block-comments-query-loop"><!-- wp:comment-template -->
+	<!-- wp:post-comment-author {"isLink":true} /-->
```
jffng, line 3: > I'm able to place it in manually via the site editor, and that generally seems to work?

I'm confused about how this is expected to work in 5.9. 

The `post-comment-author`, `post-comment-date`, `post-comment-content`,  and `post-comments-form` are all [currently marked as experimental](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/index.js#L245-L257).

jffng: Can we use blocks that are marked as experimental, they just don't show up in the inserter?

kjellr: No, that doesn't sound right.

jffng: Left a comment here to get some clarification: https://github.com/WordPress/gutenberg/pull/35979#issuecomment-956540014

michalczaplinski:
Hi folks, I've noted [here](https://github.com/WordPress/gutenberg/pull/35979#issuecomment-956681949) that I don't think we should use the Comments Query Loop block in the twentytwentytwo as it's still very basic and not ready to be included in 5.9.

jffng:
I think we can close this since `post-comments` will now be used in 5.9: https://github.com/WordPress/gutenberg/pull/36132

Thanks for pushing the conversation forward with the PR @scruffian.

PR #195, opened 2021-11-02
Tidies up functions.php to aid readability

### LINKED ISSUE #191: Better organize the code in functions.php
There's not actually much code in functions.php, but it is very dense. We should clean this up to so that it's easier to read with some extra space and file/function headers. 

Here's our functions.php vs. the one for Twenty Twenty-One for example:

Twenty Twenty-Two|Twenty Twenty-One
---|---
Screen Shot 2021-11-01 at 12 33 32 PM|Screen Shot 2021-11-01 at 12 33 43 PM

mburridge:
**Description**

Cleans up` functions.php` to so that it's easier to read and adds some extra space and file/function headers.

As per [#191 ](https://github.com/WordPress/twentytwentytwo/issues/191).

**Screenshots**

N/A

**Testing Instructions** 

N/A Only line spacing and comment blocks added

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -12,16 +12,16 @@
 
 if ( ! function_exists( 'twentytwentytwo_support' ) ) :
 
-	/** 
+	/**
 	 * Add support for core block visual styles.
 	 * Styles load in both the editor and the front end.
-	 * 
+	 *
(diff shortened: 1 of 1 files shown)


jffng (approved):
LGTM.

PR #196, opened 2021-11-02
Hide 404 pattern from inserter

jffng:
**Description**

This PR hides the 404 pattern from the block inserter. I'm not actually sure if we want to do this, but thought I'd open a PR to discuss.

**Testing Instructions** 

- Open the pattern inserter and go to the General patterns
- Verify the 404 pattern does not show up


inc/patterns/general-404.php
```diff
@@ -15,4 +15,5 @@
 					<!-- wp:spacer {"height":128} -->
 					<div style="height:128px" aria-hidden="true" class="wp-block-spacer"></div>
 					<!-- /wp:spacer -->',
+	'inserter'   => false,
```
kjellr: Really minor, but can we move this up above the content? I think it's less likely to be missed that way. Also, do we still need the category attribute?

jffng: > Also, do we still need the category attribute?

Nope!

> can we move this up above the content?

👍

kjellr:
This seems ok, but what do you think about creating a new `hidden-` filename prefix for this and for the one that'll likely land in #183? It seems a little weird to name them `general-` and `header-` etc. 

So this one might become just `hidden-404.php`?

kjellr (approved):
Looks good! Thanks.

Issue #91, opened 2021-10-12
Footer Patterns: Followup Fixes


kjellr:
Following #85, there are a couple small things that we should try and readdress for the footer patterns:

## Footer with query, title, and citation

- [x] Title and citation are oddly aligned on small screens. ([Screenshot](https://cloudup.com/ciMq4alVqwh))

## Footer with query, featured images, title, and citation

- [x] Title and citation are oddly aligned on small screens. ([Screenshot](https://cloudup.com/c7074k8_SBK))

## Dark footer with title and citation

- [ ] Title and citation have no left/right padding on small screens. ([Screenshot](https://cloudup.com/clwecooXWRw))

carolinan:
I did not find a great solution for the dark footer with title and citation. Adding padding to the sides and not only top and bottom would mean the texts don't align with the content.
One alternative would be to center the texts.

PR #199, opened 2021-11-03
[Pattern] Update Footer with query, title, and citation

carolinan:
**Description**
Partial for https://github.com/WordPress/twentytwentytwo/issues/91
Solves the alignment of the powered by text on small screens.
It does this by removing the right alignment from the paragraph, and adding a second row block.

On larger screen sizes, the texts are still aligned to the left and right respectively thanks to the space-between justification.

It also solves an issue I noticed from the screenshot below; when a post has no title, the post could not be reached.
To solve that I enabled the link option on the post date block.

**Screenshots**
The footer text is aligned vertically

Animation showing how the footer text position changes depending on screen width

**Testing Instructions** 
1. Add the "Footer with query, title, and citation" block pattern in the site editor. 
2. View the mobile preview, and view the front on different screen sizes.

### DIFF
--- a/inc/patterns/footer-query-title-citation.php
+++ b/inc/patterns/footer-query-title-citation.php
@@ -13,7 +13,7 @@ return array(
 
 					<!-- wp:post-excerpt /-->
 
-					<!-- wp:post-date /-->
+					<!-- wp:post-date {"isLink":true} /-->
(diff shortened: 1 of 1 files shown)


kjellr (approved):
Great idea! Thanks for solving this, @carolinan!

Issue #101, opened 2021-10-13
File block download button needs styles


jffng:
**Describe the solution you'd like**
The file block's download button is not styled: 

<img width="404" alt="Screen Shot 2021-10-13 at 9 41 11 AM" src="https://user-images.githubusercontent.com/5375500/137144588-20f06515-8717-46c7-86d4-513db5d9cd65.png">

The button should match the theme's default button styles. Can button styling of the file block be done with theme.json yet?

MaggieCabrera:
> The button should match the theme's default button styles. Can button styling of the file block be done with theme.json yet?

AFAIK no, it would fall under this: https://github.com/WordPress/gutenberg/issues/29167

kjellr:
Fixed by #235.

Issue #11, opened 2021-10-05
Search block is missing styles


kjellr:
Current state:

<img width="680" alt="Screen Shot 2021-10-05 at 1 19 12 PM" src="https://user-images.githubusercontent.com/1202812/136071809-f5d1ed67-a581-4dfd-864a-13a140635cfb.png">

MaggieCabrera:
I had the markup for testing this block around, I'm dropping it here for easier testing:

```
<!-- wp:search {"label":"Search","placeholder":"Placeholder text...","buttonText":"Search"} /-->

<!-- wp:search {"label":"Search","showLabel":false,"placeholder":"Placeholder text...","buttonText":"Search"} /-->

<!-- wp:search {"label":"Search","placeholder":"Placeholder text...","buttonText":"Search","buttonPosition":"button-inside"} /-->

<!-- wp:search {"label":"Search","showLabel":false,"placeholder":"Placeholder text...","buttonText":"Search","buttonPosition":"button-inside"} /-->

<!-- wp:search {"label":"Search","placeholder":"Placeholder text...","buttonText":"Search","buttonUseIcon":true} /-->

<!-- wp:search {"label":"Search","showLabel":false,"placeholder":"Placeholder text...","buttonText":"Search","buttonUseIcon":true} /-->

<!-- wp:search {"label":"Search","placeholder":"Placeholder text...","buttonText":"Search","buttonPosition":"button-inside","buttonUseIcon":true} /-->

<!-- wp:search {"label":"Search","showLabel":false,"placeholder":"Placeholder text...","buttonText":"Search","buttonPosition":"button-inside","buttonUseIcon":true} /-->

<!-- wp:search {"label":"Search","placeholder":"Placeholder text...","buttonText":"Search","buttonPosition":"no-button"} /-->

<!-- wp:search {"label":"Search","showLabel":false,"placeholder":"Placeholder text...","buttonText":"Search","buttonPosition":"no-button"} /-->
```

How it looks like on TT2:

<img width="757" alt="Screenshot 2021-10-06 at 19 36 05" src="https://user-images.githubusercontent.com/3593343/136254955-23f2b43a-12f9-4654-a6fd-a0022c7754b0.png">

carolinan:
Is the correct style for the search block the one in [the Figma template for the 404 page?](https://www.figma.com/file/76mfUcaK4QDlrXElk8MK3H/Twenty-Twenty-Two?node-id=815%3A4945)

[...]

PR #202, opened 2021-11-03
Try styling Search and File buttons

jffng:
**Description**

This PR tries to style the Search and File blocks' buttons.

Since there is not yet a way to style those blocks via theme.json / global styles, the PR links the relevant styles to the Button block's using [this idea](https://github.com/WordPress/gutenberg/pull/36130#issuecomment-957360547). 

Attempts to solve #101 and #11.

**Screenshots**

Before | After
------- | ------
Screen Shot 2021-11-03 at 11 09 02 AM | Screen Shot 2021-11-03 at 11 08 55 AM

**Testing Instructions** 

1. Create a post and add this markup: https://gist.github.com/jffng/263436360988105ab73c35effd494fe0 
3. Verify the colors and border match the Button block's in the editor and front-end
4. Go to the site editor global styles panel and customize the button block's colors. Select a _non-palette_ color*
5. Verify the colors have changed in the post

Note that this currently breaks if you select a preset color because of the value of the variable is not valid CSS, so we'd have to parse / transform it in the theme. This is what ends up getting output when you select a preset (see the background color):

Screen Shot 2021-11-03 at 11 13 39 AM

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -45,6 +45,9 @@ if ( ! function_exists( 'twentytwentytwo_styles' ) ) :
 		// Add styles inline.
 		wp_add_inline_style( 'twentytwentytwo-style', twentytwentytwo_get_font_face_styles() );
 
+		// Add global styles linked variables.
+		wp_add_inline_style( 'twentytwentytwo-style', twentytwentytwo_theme_variables() );
+
 		// Add metadata to the CSS stylesheet.
 		wp_style_add_data( 'twentytwentytwo-style', 'path', get_template_directory() . '/style.css' );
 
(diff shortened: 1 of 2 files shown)


functions.php
```diff
@@ -56,6 +59,22 @@ function twentytwentytwo_styles() {
...
+function twentytwentytwo_theme_variables() {
+	if ( ! function_exists( 'gutenberg_get_global_styles' ) ) {
+		return;
+	}
+
+	// Get the properties of the button that we want to use in other blocks like search and file.
+	$button_styles = gutenberg_get_global_styles( array(), 'core/button' );
+
+	// Create variables that point to the value of those customizable properties.
+	return ':root, body { --wp--theme--button--color--text: ' . $button_styles['color']['text'] . '; --wp--theme--button--color--background: ' . $button_styles['color']['background'] . '; --wp--theme--button--border--radius: ' . $button_styles['border']['radius'] . '; }';
```
jffng, line 76: FWIW I'm hesitant to add a workaround like this to the theme. Open to other opinions and feedback.

kjellr: Yeah, this feels just a bit too complicated for my liking. Since this is the sort of thing that should most definitely be handled natively by Gutenberg at some point, I lean towards not patching it in the theme. 😕

I'm not too concerned about the file block, but the Search block seems like a higher priority since we're actually using that in a template.

kjellr:
Closing in favor of #235.

Issue #206, opened 2021-11-04
Visible horisontal scroll bar


carolinan:
**Describe the bug**
At some screen sizes there is a visible horisontal scroll bar.

**To Reproduce**
1. Go to the front of the website
2. Reduce the browser window width to somewhere between 800 and 900px.
3. See the scroll bar

**Expected behavior**
I expect there to not be a visible horisontal scroll bar.

**Screenshots**
![visible scroll bar at the bottom of the browser window](https://user-images.githubusercontent.com/7422055/140287162-e886f021-73e5-4763-8bb0-164465901247.png)

**Device:**
 - Device: PC
 - OS: Windows10
 - Browser: Chrome

**Additional context**
I believe there was a discussion in the Gutenberg GitHub repository about a default "reset" CSS for themes, but 
I can't find it now and whatever the result was, it did not include overflow.
Considering adding `overflow-x: hidden;` to the body element for the front only, in style.css

kjellr:
@carolinan I'm unable to reproduce this, using the same demo content. Are you able to pinpoint the blocks that are causing this?

carolinan:
I believe it is the submenu. And looking at the screenshot above, the rightmost menu item has a sub menu.

![scroll](https://user-images.githubusercontent.com/7422055/140317297-1cff7f04-278f-4a35-9a27-5c65a707d12e.gif)

kjellr:
Thanks! @jasmussen: as a navigation block expert, is this something you have any insight into? 

Since the theme isn't providing any of its own styling, we shouldn't need to fix this on the theme end.

jasmussen:
I haven't found a good way to share test markup with the recent changes, so I'll try to speak in a bit more general terms about how things _should_ work.

- When submenus are closed, their contents should not take up space which could cause scrollbars. 
- When a menu justified left, submenus should open to the right. When justified right, they should open to the left.

At the moment we have no further positioning smarts for which direction a submenu should open in, which means there's a little bit of editorial responsibility present. We also don't have handling in place to ensure endless submenus flipping back and forth to avoid scrollbars — it's complex javascript we'll likely need to address in the future, but for the time being the submenu directional smarts are based on justifications.

However from the screenshots and GIFs (thank you), it appears that the menu _is_ justified right in the first screenshot, yet still opens rightward when it should open left. It even looks as though it takes up space while closed. Both of those would be regressions. I'll take a look, and in the mean time, I hope the above makes sense.

[...]

PR #207, opened 2021-11-04
Don't inline style.css, but enqueue it.

carolinan:
**Description**
Register and enqueue style.css without printing it inline.

Right now, the style.css file isn't actually used, it is empty. It was enqueued this way to allow adding CSS to the front,
for https://github.com/WordPress/twentytwentytwo/issues/206 and any other CSS _we are not sure will be solved by theme.json in time._

**Testing Instructions** 
1. View source.
2. Confirm that the style.css file header is not printed in the source.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -40,13 +40,15 @@ if ( ! function_exists( 'twentytwentytwo_styles' ) ) :
 	function twentytwentytwo_styles() {
 
 		// Register theme stylesheet.
-		wp_register_style( 'twentytwentytwo-style', '' );
+		wp_register_style(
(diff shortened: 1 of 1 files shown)


jffng (commented):
I'm good with this change. Should we add it to the editor while we're here too?

carolinan:
It depends. Perhaps wait until there is CSS that needs to be added?
About to test, but it looks like it wont be needed for https://github.com/WordPress/twentytwentytwo/issues/206

jffng (approved):

PR #209, opened 2021-11-04
Try using the site padding control

kjellr:
For testing against. https://github.com/WordPress/gutenberg/pull/36214

This PR: 

- Removes all left/right padding from the site templates.


theme.json
```diff
@@ -341,7 +341,8 @@
 			}
 		},
 		"spacing": {
-			"blockGap": "1.5rem"
+			"blockGap": "1.5rem",
+			"padding": "0 var(--wp--custom--spacing--small)"
```
youknowriad: This should be an object with left, right, top, bottom and not a string.

youknowriad: It should be moved to `settings.layout.padding` as well.

kjellr: Ah, ok that gets things working right! Thanks.

kjellr:
I'm going to close this one for now, since it's outdated and the corresponding PR isn't going to be a part of 5.9.

PR #215, opened 2021-11-08
Fix php compatibility and linting errors.

jffng:
**Description**

This PR removes a trailing comma in `functions.php` that I think is causing a test to fail in the core PR: https://github.com/WordPress/wordpress-develop/runs/4142008741?check_suite_focus=true 

It also fixes a couple standards errors.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -44,7 +44,7 @@ if ( ! function_exists( 'twentytwentytwo_styles' ) ) :
 			'twentytwentytwo-style',
 			get_template_directory_uri() . '/style.css',
 			array(),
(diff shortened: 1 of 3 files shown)


jffng:
Let's wait to use this PR to compile all the changes that are needed to get the theme initially accepted to Core trunk.

jeffikus (approved):
@jffng this LGTM, it passes the checks as well. I've seen some strange phpcs fails on multiline array before, so keep an eye out for that when you merge this. Otherwise this LGTM!

aristath (approved):

PR #216, opened 2021-11-08
Update contributors list.

jffng:
**Description**

Updates the contributors list. Might have missed some folks but I went through the issues and credited from there as well.


kjellr (approved):
Thanks for pulling this together!

Issue #210, opened 2021-11-05
Switch theme.json to v2 (remove "custom" prefix)


carolinan:
While version 1 will still work in 5.9, I think the theme should use version 2:
https://github.com/WordPress/gutenberg/pull/36155

Do we also need to wait for https://github.com/SchemaStore/schemastore/pull/1907?

carolinan:
@ajlende is this the correct way to add the schema back?
`"$schema": "https://schemas.wp.org/trunk/theme.json",`
Does it now support version 2 of theme.json?

(I needed to remove it here: https://github.com/WordPress/twentytwentytwo/pull/217/files#diff-160c735a986cf8c05e1f6b2e9ac89dd0410562d2fe5ccbdc93157bd92f3dc2bcL2)

ajlende:
That is correct for Visual Studio Code for the latest version. The latest version, however, had a bug where the version was not updated with the changed properties, so I added a fix here https://github.com/WordPress/gutenberg/pull/36917

PR #217, opened 2021-11-09
Update theme.json to v2

carolinan:
**Description**
Removes custom prefixes.
Temporarily removes schema since the schema only supports version 1.
Also changed some spaces to tabs.

**Testing Instructions** 
1. Activate the theme.
2. Confirm that line height, border, padding and margin still works.


aristath (approved):

kjellr (approved):
Thanks @carolinan! 

Fixes #210.

PR #218, opened 2021-11-09
Make "ABOUT US" use the Gutenberg text case setting

kjellr:
Addresses an issue raised in https://github.com/WordPress/wordpress-develop/pull/1817#discussion_r745039417.

I'd like to keep this uppercase for text hierarchy reasons, but we'll switch to using the Gutenberg text case control. 

If that is a problem for other languages, presumably Gutenberg should be able to handle that elegantly itself (Otherwise this setting is problematic for all themes).


jffng:
Addressed in #223.

Issue #161, opened 2021-10-25
Opt the theme into font smoothing for better font rendering


kjellr:
This may require a bit of CSS, unless https://github.com/WordPress/gutenberg/issues/35934 ends up happening. 

```
-moz-osx-font-smoothing: grayscale;
-webkit-font-smoothing: antialiased;
```

kjellr:
Here's a before/after comparison:

_Without font smoothing_

<img width="1056" alt="Screen Shot 2021-10-25 at 3 45 36 PM" src="https://user-images.githubusercontent.com/1202812/138760318-d481edcd-7660-4b57-b75b-7e379c91a942.png">

_With font smoothing_

<img width="1082" alt="Screen Shot 2021-10-25 at 3 45 53 PM" src="https://user-images.githubusercontent.com/1202812/138760330-77356898-649a-420b-a385-a4feab07802f.png">

bgardner:
As I mentioned [here](https://github.com/WordPress/gutenberg/issues/35934#issuecomment-954784889), I am a huge advocate of font-smoothing. Especially for TT2, since it’s absence is the first thing I notice every time I view the demo or test it locally.

sagarnasit:
Based on [this discussion](https://github.com/WordPress/gutenberg/pull/36092#issuecomment-957401611) font smoothing is supposed to be added via theme style rather than theme.json.

Let me know if I can help with that.

PR #220, opened 2021-11-09
Add font smoothing + text link styles

### LINKED ISSUE #108: Text links need improved styles
From [the mockups](https://www.figma.com/file/76mfUcaK4QDlrXElk8MK3H/Twenty-Twenty-Two?node-id=729%3A4866): 

---

image

---

- Links need more refined text-decoration styles, so that the thin font isn't paired with a [horribly thick underline](https://cloudup.com/czdjAkphnKr). (I think we could probably elevate this issue from "enhancement" to a "bug" because of this bullet). 
- Link underlines should become dotted on hover and in their active state.  
- Focus state should be dotted.
- The mockups indicate that large post titles should _not_ have underlines by default, but that their hover and active states should have a solid underline.

May end up being related to: #99 
Upstream issue: https://github.com/WordPress/gutenberg/issues/27075

kjellr:
Fixes #108 and #161. 

Loads the `style.css` file in the editor, and then adds CSS for font smoothing, and for basic text link styles (including hover, active, and focus). These styles do not impose a `text-decoration: underline` rule, so they won't effect or override elements that don't have an underline by default (like the button block).

Each group of rules is accompanied by a description of why it's implemented with CSS, and points to a Gutenberg issue for more details. 

## Screenshots

_Link hover state:_
 
link

Note: I went with a dashed instead of dotted underline so that it better matches the dashed lines used throughout the bird patterns. 

Before|After
---|---
Screen Shot 2021-11-09 at 10 28 52|Screen Shot 2021-11-09 at 10 28 40


kjellr:
Note that some of the underline rules can be added to theme.json [via this technique](https://github.com/WordPress/twentytwentytwo/issues/108#issuecomment-960503784), but it seems a little delicate to me, and we'd still require CSS for some other text link styles.

functions.php
```diff
@@ -13,17 +13,17 @@
...
 	 *
 	 * @return void
 	 */
 	function twentytwentytwo_support() {
 
+		// Add support for block styles.
 		add_theme_support( 'wp-block-styles' );
 
+		// Enqueue editor styles.
+		add_editor_style( 'style.css' );
```
jffng: Is it safer to do this?

```suggestion
		add_editor_style( get_template_directory_uri() . '/style.css' );
```

style.css
```diff
@@ -16,3 +16,34 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
...
+ */
+
+body {
+	-moz-osx-font-smoothing: grayscale;
+	-webkit-font-smoothing: antialiased;
+}
+
+/*
+ * Text link styles.
+ * Necessary until the following issue is resolved in Gutenberg: ;
```
jffng:
```suggestion
 * Necessary until the following issue is resolved in Gutenberg:
```

style.css
```diff
@@ -16,3 +16,34 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
...
+}
+
+a:hover,
+a:focus {
+	text-decoration-style: dashed;
+}
+
+a:active {
+	text-decoration: none;
+}
```
jffng: Every file should end in a newline. 

```suggestion
}

```

jffng (approved):
Some small changes, otherwise LGTM.

functions.php
kjellr: Turns out this doesn't work. 😄 I'll remove in a subsequent commit.

PR #223, opened 2021-11-10
Sync with WP `trunk`

desrosj:
**Description**

This PR contains all the changes made during pre-commit review when merging into WordPress Core.

Discussion for these can be found in https://github.com/WordPress/wordpress-develop/pull/1817.


jffng:
Summarizing the changes for good measure:

- Ensures every file ends with a newline
- Fixes trailing comma in functions.php (#215)
- Changes all instances of uppercase text to use text-transform instead (#218)
- Fixes doc block for all function definitions
- Removes `function_exists` calls for block patterns since the theme won't work with anything earlier than 5.9, and those functions were introduced in 5.5
- Adds an php code opening to index.php to pass tests
- Bumps version and changelog to 1.0

jffng (approved):

PR #235, opened 2021-11-15
Try adding simple search + file button styles

kjellr:
Alternate to #202. 

This takes a purely CSS approach to styling the file and search block buttons. 

Something like this is required until the following Gutenberg issues are resolved:
- https://github.com/WordPress/gutenberg/issues/27760
- https://github.com/WordPress/gutenberg/issues/36444

Test data: 
https://gist.github.com/jffng/263436360988105ab73c35effd494fe0

Before|After
---|---
Screen Shot 2021-11-15 at 12 21 21 PM|Screen Shot 2021-11-15 at 12 20 33 PM


functions.php
```diff
@@ -25,7 +25,7 @@ function twentytwentytwo_support() {
 		add_theme_support( 'wp-block-styles' );
 
 		// Enqueue editor styles.
-		add_editor_style( get_template_directory_uri() . '/style.css' );
+		add_editor_style( 'style.css' );
```
kjellr, line 28: This undoes the commit here: https://github.com/WordPress/twentytwentytwo/pull/220#discussion_r745795293

In my testing, this prevented the stylesheet from loading in the editor.

style.css
```diff
@@ -35,8 +35,8 @@ body {
  */
 
 a {
-    text-decoration-thickness: 1px;
-    text-underline-offset: 0.25ch;
+	text-decoration-thickness: 1px;
+	text-underline-offset: 0.25ch;
```
kjellr, line 39: Changing spaces to tabs. 😉

style.css
```diff
@@ -47,3 +47,24 @@ a:focus {
...
+ */
+
+.wp-block-search__button,
+.wp-block-file .wp-block-file__button {
+	background-color: var(--wp--preset--color--primary);
+	border-radius: 0;
+	border: none;
+	color: var(--wp--preset--color--background);
+	font-size: var(--wp--preset--typography--font-size--normal);
+	padding: calc(.667em + 2px) calc(1.333em + 2px);
```
kjellr, line 65: We could leave this padding line out to keep this even more minimal — it makes this mirror the normal button padding, but I'm a little indifferent about it.

style.css
```diff
@@ -47,3 +47,24 @@ a:focus {
...
+}
+
+.wp-block-file a.wp-block-file__button:hover {
+	opacity: 1;
+}
```
jffng:
```suggestion
}

```

functions.php
jffng: I can't recreate this but I'm fine with the change, based on the [docs](https://developer.wordpress.org/reference/functions/add_editor_style/#more-information) for this function.

style.css
jffng: I think let's keep it.

style.css
```diff
@@ -47,3 +47,24 @@ a:focus {
...
+ * https://github.com/WordPress/gutenberg/issues/36444
+ * https://github.com/WordPress/gutenberg/issues/27760
+ */
+
+.wp-block-search__button,
+.wp-block-file .wp-block-file__button {
+	background-color: var(--wp--preset--color--primary);
+	border-radius: 0;
+	border: none;
+	color: var(--wp--preset--color--background);
```
jffng, line 63: What do you think about adding this line: 

```suggestion
	color: var(--wp--preset--color--background);
	cursor: pointer;
```

jffng (approved):
Some minor comments but LGTM.

style.css
kjellr: I recall there being a major discussion around this cursor behavior at some point in Core. I think we should leave it to be the default behavior for right now — if that gets changed it should happen in Gutenberg.

PR #238, opened 2021-11-15
Add filters for block patterns and block pattern categories

### LINKED ISSUE #219: Allow block patterns to be filtered
It was suggested to add a filter that would allow child theme authors to remove block patterns.

More context here and suggestions here: https://github.com/WordPress/wordpress-develop/pull/1817#discussion_r745036953

cc @desrosj

desrosj:
**Description**

This adds two new filters for adjusting the theme's block pattern categories and block patterns, respectively.
- `twentytwentytwo_block_pattern_categories`
- `twentytwentytwo_block_patterns`.

This makes it possible for a site owner to easily adjust the list of available block patterns and pattern categories without having to override the entire function, which prevents all future changes to this function from being available when this happens).

Fixes #219.

### DIFF
--- a/inc/block-patterns.php
+++ b/inc/block-patterns.php
@@ -5,127 +5,125 @@
  * @since Twenty Twenty-Two 1.0
  */
 
-if ( ! function_exists( 'twentytwentytwo_register_block_patterns' ) ) :
+/**
+ * Registers block patterns and categories.
+ *
+ * @since Twenty Twenty-Two 1.0
+ */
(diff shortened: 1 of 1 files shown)


jffng (approved):
I think this is a welcome change for this reason: 

> This makes it possible for a site owner to easily adjust the list of available block patterns and pattern categories without having to override the entire function, which prevents all future changes to this function from being available when this happens).

Thanks @desrosj!

PR #240, opened 2021-11-16
Preload main web font to improve LCP performance

### LINKED ISSUE #239: Improve LCP performance by preloading web font
Today I ran some performance tests for Twenty Twenty-Two, to review how it is performing overall and specifically in comparison with the previous default theme Twenty Twenty-One, which is particularly interesting also to get an early idea on performance implications between the traditional theming approach and full site editing.

The great thing first: Twenty Twenty-Two does a much better job in terms of CSS and JavaScript. Particularly for CSS, a lot less CSS is loaded, particularly because CSS is now injected on demand rather than loading a massive CSS file which contains all CSS of the theme, resulting in a lot of unused CSS which is wasteful. JavaScript is slightly less as well, and it's great to see that the theme doesn't even (need to) bundle any JavaScript itself. The full site editing paradigm shows a clear benefit here. 🎉 

Now here's the problem: Twenty Twenty-Two currently loads a web font (which Twenty Twenty-One doesn't), and because of that the overall [Largest Contentful Paint metric](https://web.dev/lcp/) is still worse than for Twenty Twenty-One, despite the above performance wins. This can be fixed though.

Of course the web font files have a considerable size which we probably cannot affect, but what we can do to improve performance is preload them. Right now, the fonts are "buried" in the CSS, so the browser first needs to parse the CSS to even detect that there is a web font to load. By adding a `preload` directive, we can tell the browser to load the font right away, so the loading process can start earlier. [See this codelab on preloading web fonts for reference.](https://web.dev/codelab-preload-web-fonts/)

I tested this fix already and it **notably improves Largest Contentful Paint, by around 10% at the median** (~2.0s with preloading vs ~2.3s without preloading in my tests). I tested it for both an archive and a single post. See the full results here:

* [Without vs with preloading (archive)](https://www.webpagetest.org/video/compare.php?medianMetric=chromeUserTiming.LargestContentfulPaint&highlightLCP=1&thumbSize=200&ival=100&end=visual&tests=211115_BiDc4Q_8e6900f1a087ff369d2188376d38668c-l:withoutpreload,211115_AiDcFT_56c73ab0bfac39ffbf78d3e222ae5981-l:withpreload)
* [Without vs with preloading (single post)](https://www.webpagetest.org/video/compare.php?medianMetric=chromeUserTiming.LargestContentfulPaint&highlightLCP=1&thumbSize=200&ival=100&end=visual&tests=211115_AiDc4M_0636f0309b063f2b6e9175363ca5c8f0-l:withoutpreload,211115_AiDcK3_4bc1caaba731fe4fed9ab4b532954cc1-l:withpreload)

felixarntz:
**Description**

This PR improves LCP performance of Twenty Twenty-Two by ~10% by preloading the main web font used. See #239 for more explanation and full before/after performance test results.

It introduces a `twentytwentytwo_preload_webfonts()` function to render the `preload` tag for the font on `wp_head`.

Fixes #239.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -119,5 +119,27 @@ if ( ! function_exists( 'twentytwentytwo_get_font_face_styles' ) ) :
 
 endif;
 
+if ( ! function_exists( 'twentytwentytwo_preload_webfonts' ) ) :
+
+	/**
+	 * Preloads the main web font to improve performance.
+	 *
+	 * Only the main web font (font-style: normal) is preloaded here since that font is always relevant (e.g. it used
+	 * on every heading). The other font is only needed if there is any applicable content in italic style, and
+	 * therefore preloading it would in most cases regress performance when that font would otherwise not be loaded at
+	 * all.
+	 *
+	 * @since Twenty Twenty-Two 1.0
+	 */
+	function twentytwentytwo_preload_webfonts() {
+		?>
+		<link rel="preload" href="<?php echo esc_url( get_theme_file_uri( 'assets/fonts/SourceSerif4Variable-Roman.ttf.woff2' ) ); ?>" as="font" type="font/woff2" crossorigin>
+		<?php
+	}
+
+endif;
+
+add_action( 'wp_head', 'twentytwentytwo_preload_webfonts' );
+
 // Add block patterns
 require get_template_directory() . '/inc/block-patterns.php';
--- a/functions.php
+++ b/functions.php
@@ -132,7 +132,7 @@ if ( ! function_exists( 'twentytwentytwo_preload_webfonts' ) ) :
 	 * all.
 	 *
 	 * @since Twenty Twenty-Two 1.0
-	 * 
+	 *
 	 * @return void
 	 */
 	function twentytwentytwo_preload_webfonts() {
--- a/functions.php
+++ b/functions.php
@@ -103,6 +103,7 @@ if ( ! function_exists( 'twentytwentytwo_get_font_face_styles' ) ) :
 			font-weight: 200 900;
 			font-style: normal;
 			font-stretch: normal;
+			font-display: swap;
 			src: url('" . get_theme_file_uri( 'assets/fonts/SourceSerif4Variable-Roman.ttf.woff2' ) . "') format('woff2');
 		}
 
@@ -111,6 +112,7 @@ if ( ! function_exists( 'twentytwentytwo_get_font_face_styles' ) ) :
 			font-weight: 200 900;
 			font-style: italic;
 			font-stretch: normal;
+			font-display: swap;
 			src: url('" . get_theme_file_uri( 'assets/fonts/SourceSerif4Variable-Italic.ttf.woff2' ) . "') format('woff2');
 		}
 		";
--- a/functions.php
+++ b/functions.php
@@ -132,6 +132,8 @@ if ( ! function_exists( 'twentytwentytwo_preload_webfonts' ) ) :
 	 * all.
 	 *
 	 * @since Twenty Twenty-Two 1.0
+	 * 
+	 * @return void
 	 */
 	function twentytwentytwo_preload_webfonts() {
 		?>


westonruter:
Alternatively (or in addition) what about adding `font-display:optional` to the two `@font-face` rules?

https://github.com/WordPress/twentytwentytwo/blob/a11fb4932a0f79dc2668e92f118fd315256be9d9/functions.php#L101-L115

If a good enough fallback `serif` font is included in the font stack, then the font could be taken out of the critical rendering path altogether.

See [Prevent layout shifting and flashes of invisible text (FOIT) by preloading optional fonts](https://web.dev/preload-optional-fonts/).

westonruter:
The lack of `font-display` is also flagged by Lighthouse:

image

functions.php
```diff
@@ -119,5 +119,27 @@ function twentytwentytwo_get_font_face_styles() {
...
+
+	/**
+	 * Preloads the main web font to improve performance.
+	 *
+	 * Only the main web font (font-style: normal) is preloaded here since that font is always relevant (e.g. it used
+	 * on every heading). The other font is only needed if there is any applicable content in italic style, and
+	 * therefore preloading it would in most cases regress performance when that font would otherwise not be loaded at
+	 * all.
+	 *
+	 * @since Twenty Twenty-Two 1.0
```
jffng, line 134:
```suggestion
	 * @since Twenty Twenty-Two 1.0
	 * 
	 * @return void
```

jffng (commented):
Thank you for taking the time to look into the performance and suggest the change @felixarntz! I gave this a spin and it's working as expected.

> what about adding font-display:optional to the two @font-face rules?

I am good with this suggestion as well, just want to make sure @kjellr understands the implication that it's possible for some text to show up un-styled and then styled when the font loads as [described here](https://web.dev/preload-optional-fonts/).

jffng:
On second thought, I don't think `font-display: optional;` is the behavior we want. If the font is not loaded before the 100ms block period, it won't display the styled text, which is not ideal: 

Screen Shot 2021-11-16 at 12 23 25 PM

I think we want `font-display: swap;` instead, @kjellr what do you think?

westonruter:
> On second thought, I don't think `font-display: optional;` is the behavior we want. If the font is not loaded before the 100ms block period, it won't display the styled text, which is not ideal:

I think this is the why [Prevent layout shifting and flashes of invisible text (FOIT) by preloading optional fonts](https://web.dev/preload-optional-fonts/) is advising a `preload` for the font which has `font-display:optional`: by loading it early, it's more likely that it will have been downloaded before the 100ms block period has transpired.

felixarntz:
What would be the benefit of `font-display:optional` over `font-display:swap`? The latter seems to make more sense to me from my understanding as well.

Let's think about this further, but it might be best to open a follow-up PR to not block this enhancement.

westonruter:
The benefit of `font-display:swap` is that the desired font will ultimately be displayed and it has no block period, so the text will appear immediately. The downside is that there will be layout shifting. With `font-display:optional` there is a very short block period, but then if the font doesn't load in time then the fallback will be displayed (and no layout shift will occur) until the page is reloaded with the cache primed.

If going with `font-display:swap`, then this should go along with `size-adjust` to reduce possible layout shift. See [CSS `size-adjust` for `@font-face`](https://web.dev/css-size-adjust/).

westonruter:
I agree this can be addressed in a follow-up PR. I'll look into it.

westonruter (approved):

westonruter:
Actually, I'm not seeing there being any size difference between `Times` (`serif`) and `Source Serif Pro`. Therefore, I think we should freely add `font-display:swap`.

In the two following examples, the network is throttled to 3G speeds.

## Without `font-display`

Notice the text is hidden for a long time here.

https://user-images.githubusercontent.com/134745/142040610-e2e425be-7903-40ec-90aa-cae46333ed95.mov

## With `font-display:swap`

Here the serif font shows up very quickly so the user can start reading.

https://user-images.githubusercontent.com/134745/142040633-b0840e75-107e-410e-a496-4356489546e1.mov

I'm not entirely sure why the `serif` font shows up momentarily in both cases. There is no visible layout shift when the fonts load.

kjellr:
👍 Yeah, `swap` seems more ideal here, so let's go with that.

felixarntz:
@kjellr @jffng @westonruter Updated the PR to include `font-display: swap` on both web fonts.

jffng (approved):
Thanks everyone.

PR #243, opened 2021-11-16
Rework the default template for a better first experience

### LINKED ISSUE #116: Add starter content

kjellr:
Potentially closes https://github.com/WordPress/twentytwentytwo/issues/116. 

This PR tackles a couple issues that came up during some recent testing: 

- The initial appearance of the theme is underwhelming — it's just a white screen. We don't have a solution to Starter Content yet, and we don't have any leads on how that experience might be improved soon. 
- The theme lacked distinct `index` and `archive` templates, which made editing the homepage index template a little awkward: In the editor you'd see an "Archive Title" block by default when you opened the Site editor, but this block was invisible on the front end. 

This PR tackles that by changing our single `index.html` template into three separate templates: 

1. `archive.html`: This is identical to today's `index.html`. It has an Archive Title block.
2. `index.html`: Same as today's `index.html`, but with no Archive Title block. This may be a useful fallback for CPTs or media, where you don't necessarily need an "Archive Title" displayed at the top. 
3. `home.html`: A new template. This uses a new "Small Dark Header" template part that is designed to bring a little bit of the personality of our `page-home.html` template into the default blog page. When users install a fresh site, this will be what they see first. This is a way to basically give them some nice "starter content" without actually giving them starter content. I think it's good enough to just close out #116. 

## Screenshots: 

(All screenshots taken on fresh install)

Before:

Frontend|Editor
---|---
Screen Shot 2021-11-16 at 15 28 44|Screen Shot 2021-11-16 at 15 28 36

After: 

Frontend|Editor
---|---
Screen Shot 2021-11-16 at 15 26 20|Screen Shot 2021-11-16 at 15 27 41


theme.json
```diff
@@ -140,7 +140,7 @@
 			"spacing": {
 				"small": "max(1.25rem, 5vw)",
 				"medium": "clamp(2rem, 8vw, calc(4 * var(--wp--style--block-gap)))",
-				"large": "8rem"
+				"large": "clamp(4rem, 10vw, 8rem)"
```
kjellr, line 143: I had to make this responsive in order to provide more reasonable spacing on small screens. All of our other values were responsive already, and it seems to work just fine in other contexts where this variable is used. 

Before|After
---|---
Screen Shot 2021-11-16 at 15 37 07|Screen Shot 2021-11-16 at 15 36 47

kjellr:
With https://github.com/WordPress/twentytwentytwo/pull/243/commits/8530d121ab7fecc10c86b12ea61394afd2538ad6, the PR also adds a new block pattern to represent the header used on `home.html`: 

Screen Shot 2021-11-16 at 3 43 37 PM

jffng (approved):
This all works for me, with one disclaimer: with recent changes to the site editor, I do not know how to access the `archive` or `index` templates from the site editor.

That said, they are looking fine on the front-end.

kjellr:
Thanks! 

> This all works for me, with one disclaimer: with recent changes to the site editor, I do not know how to access the archive or index templates from the site editor.

Yeah, my understanding is that either: the template browser sidebar will be added back into 5.9, or the there will be a resolution to https://github.com/WordPress/gutenberg/issues/35994 in time. 👍

PR #247, opened 2021-11-17
Fix header.html markup

kjellr:
The latest changes to `header.html` have a little typo in them that's causing a markup error in the editor:

Before|After
---|--
Screen Shot 2021-11-16 at 7 15 13 PM|Screen Shot 2021-11-16 at 7 14 38 PM


MaggieCabrera (approved):
LGTM!

Issue #233, opened 2021-11-15
Question: OK to have multiple H1 headings per page?


Sandstromer:
**Description**
I noticed the Site Title in the header is a H1 heading, and also the Query Title in the index template is also a H1 heading.

When viewing an archive/category/author etc there are two H1 headings on the page.

The Post Title blocks within the Query loop in the archive are H2.
If the Query Title was changed to H2, then the Query Loop Post Title's should be changed to H3?

This would be OK for viewing archives, but then the main home index would be missing a H2 Query Title, and would go from Site Title H1 to the Query Loop Post Title's H3.

Hope I've managed to explain my thinking okay here, and not sure whether I'm fussing over something that should not really be of concern?

The Site Title could be changed to paragraph but I guess that would not be deemed correct?

kjellr:
In general, we shouldn't. Oftentimes, the way this works for themes is: 

- The Site Title is a `h1` on the homepage, or any page without a distinct title. 
- The Site Title is a `p` on pages where we expect there to be a distinct H1. 

Since these are pulling from the same `header.html` template part, we can't have it dynamically change that way. For now, it's calling the Site Title a `h1` everywhere, because this way we know we'll always have at least one H1 on the page. We _could_ change it to a `p`, but then I imagine we'll have some pages where there is no `h1` at all (like the `index.html` page when it's used for Blog Posts). 

If anyone has suggestions, please share!

Related: https://github.com/WordPress/twentytwentytwo/issues/82

kafleg:
These days Google doesn't mind if any page has multiple H1 tags, but standard practice is to make sure H1 is assigned to the main title of that page.

Reference: https://www.youtube.com/watch?v=zyqJJXWk0gk

kjellr:
We had [a discussion about this in #core-themes the other day](https://wordpress.slack.com/archives/C02RP4VMP/p1636990136197100?thread_ts=1636989618.188800&cid=C02RP4VMP) (signup required to view the link). I'm going to surface some of the discussion to add to what's here in the ticket: 

- WCAG Guidelines do not prohibit more than one H1 per page. 
- The `accessibility-ready` guidelines for the Theme Directory are not strictly the WCAG guidelines though. [Those guidelines _do_ require no more than one H1 per page](https://make.wordpress.org/themes/handbook/review/accessibility/required/#headings). 

We need to follow those guidelines, so unless those are changing, I suggest we: 

[...]

PR #249, opened 2021-11-18
Having an h1 on single posts

SGr33n:
The single.html template doesn't have an h1 that should be the post title.


kjellr:
Since the Site Title is a H1, implementing this (or #250) would unfortunately break compliance with [the accessibility guidelines for themes](https://make.wordpress.org/themes/handbook/review/accessibility/required/#headings). Let's continue to discuss in #233.

kjellr:
As per the discussion in https://github.com/WordPress/twentytwentytwo/issues/233#issuecomment-985743555, we should be ok using more than one H1 per page. I've updated the "No Separators" template too, and will merge this in. Thanks for the help, @SGr33n!

PR #250, opened 2021-11-18
Having h1 titles on pages

SGr33n:
The theme is using an H2 to display the page titles. That title should have an H1 tag.


kjellr:
Thanks for the issue. I'm going to combine this with #249, since they're so related.

kjellr:
Oh, sorry  — I didn't realize this was a PR. I'll leave it open for now while we discuss.

kjellr:
As per the discussion in https://github.com/WordPress/twentytwentytwo/issues/233#issuecomment-985743555, we should be ok using more than one H1 per page. I've updated the "No Separators" template too, and will merge this in. Thanks for the help, @SGr33n!

PR #252, opened 2021-11-18
Tidy up patterns that use the "bird on salmon" images

kjellr:
This PR does a couple things: 

- It makes it so that we don't need two separate "Bird on Salmon" images — we can just use the same image for both.
- It improves the responsive behavior of the "About page with media on the left" pattern by using a `min()` padding rule, like we already do for the "About page with media on the right" pattern. 

The only visible change will be for the "About page with media on the left" pattern. The graphic of the bird is no longer centered vertically in the preview, but it generally ends up being cropped in a decent way. I think this is an improvement, especially considering how much better this looks on mobile. 

Before|After
---|---
Screen Shot 2021-11-18 at 3 49 31 PM|Screen Shot 2021-11-18 at 3 48 42 PM

Before|After
---|---
Screen Shot 2021-11-18 at 3 49 54 PM|Screen Shot 2021-11-18 at 3 47 59 PM


MaggieCabrera (approved):
This LGTM. The About pattern is a bit more obscure in terms of customizing it from the editor but it's still a viable solution and I don't think should block this.

kjellr:
Thanks for the review! I woke up with the thought that I should see if one of our other images will crop better in this pattern instead. Going to give it one more try before we merge. 👍

kjellr:
Ok, I tried that other idea out, and it works but I prefer the elegance of the bird graphic we're using here. I tested it against a wide variety of typical device sizes, and the crop is generally acceptable. Let's go with this one. Thanks for the review, @MaggieCabrera!

PR #253, opened 2021-11-19
Add the remaining header template part to theme.json

carolinan:
**Description**
Add the reamining header template part to theme.json.

**Screenshots**

Before and after:
two views of the template part list in the site editor, the left image shows the name of the template part, the right image shows the title of the template part.

**Testing Instructions** 
1. With Gutenberg trunk active. open the list of template parts in the site editor.
2. See that the name of the template part is the added title.

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -356,12 +356,12 @@
 		},
 		{
 			"name": "header-large-dark",
-			"title": "Header — Dark",
+			"title": "Header (Dark, large)",
 			"area": "header"
 		},
 		{
 			"name": "header-small-dark",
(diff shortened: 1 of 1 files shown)


kjellr:
Thanks for catching this! I missed this on my PR. 

Let's rename these to "Header (Dark, Large)" and "Header (Dark, Small)". 

We're already using parentheses in this way in a couple of the patterns, so I think it makes sense to stick to that instead of the dash:
 
Screen Shot 2021-11-19 at 8 20 22 AM

jffng (approved):

PR #254, opened 2021-11-19
Add missing text domains and minor related adjustments

carolinan:
**Description**
Adds missing text domains.
Replace `esc_html__` with `esc_attr__` inside alt attribute.
Adds translator context to the "TBD" abbreviation.

### DIFF
--- a/inc/patterns/page-about-media-right.php
+++ b/inc/patterns/page-about-media-right.php
@@ -6,14 +6,14 @@ return array(
 	'title'      => __( 'About page with media on the right', 'twentytwentytwo' ),
 	'categories' => array( 'twentytwentytwo-pages' ),
 	'content'    => '<!-- wp:media-text {"align":"full","mediaPosition":"right","mediaLink":"' . esc_url( get_stylesheet_directory_uri() ) . '/assets/images/bird-on-black.jpg","mediaType":"image","style":{"elements":{"link":{"color":{"text":"var:preset|color|background"}}}},"backgroundColor":"foreground","textColor":"background"} -->
-				<div class="wp-block-media-text alignfull has-media-on-the-right is-stacked-on-mobile has-background-color has-foreground-background-color has-text-color has-background has-link-color"><figure class="wp-block-media-text__media"><img src="' . esc_url( get_stylesheet_directory_uri() ) . '/assets/images/bird-on-black.jpg" alt="' . esc_attr__( 'An image of a bird flying' ) . '"/></figure><div class="wp-block-media-text__content"><!-- wp:spacer {"height":32} -->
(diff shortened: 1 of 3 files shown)


carolinan:
Do we also need to replace  ` ’` as in  `Oh hello. My name’s Angelo`, with `&#8217;` ?

inc/patterns/page-layout-image-and-text.php
```diff
@@ -10,18 +10,18 @@
 					<h2 class="alignwide" style="font-size:clamp(4rem, 8vw, 7.5rem);font-weight:300;line-height:1.15">' . wp_kses_post( __( '<em>Watching Birds </em><br><em>in the Garden</em>', 'twentytwentytwo' ) ) . '</h2>
 					<!-- /wp:heading --></div>
 					<!-- /wp:group -->
-
```
carolinan, line 13: This spacing change is not meant to be here 🤔

inc/patterns/page-about-media-right.php
```diff
@@ -6,14 +6,14 @@
...
 					<div style="height:32px" aria-hidden="true" class="wp-block-spacer"></div>
 					<!-- /wp:spacer -->
 					<!-- wp:site-logo {"width":60} /-->
 
 					<!-- wp:group {"style":{"spacing":{"padding":{"right":"min(8rem, 5vw)","top":"min(20rem, 20vw)"}}}} -->
 					<div class="wp-block-group" style="padding-top:min(20rem, 20vw);padding-right:min(8rem, 5vw)"><!-- wp:heading {"style":{"typography":{"fontWeight":"300","lineHeight":"1.115","fontSize":"clamp(3rem, 6vw, 4.5rem)"}}} -->
-					<h2 style="font-size:clamp(3rem, 6vw, 4.5rem);font-weight:300;line-height:1.115"><em>Emery<br>Driscoll</em></h2>
+					<h2 style="font-size:clamp(3rem, 6vw, 4.5rem);font-weight:300;line-height:1.115"><em>' . esc_html__( 'Emery', 'twentytwentytwo' ) . '<br>' . esc_html__( 'Driscoll', 'twentytwentytwo' ) . '</em></h2>
```
MaggieCabrera: Could this be `' . wp_kses_post( __( 'Emery<br>Driscoll', 'twentytwentytwo' ) ) . '` instead?

jffng:
> Do we also need to replace ’ as in Oh hello. My name’s Angelo, with &#8217; ?

What would be the reason for doing this? The current way results in the correct right singular apostrophe.

jffng (commented):
Pending @MaggieCabrera's comment, this LGTM 👍

inc/patterns/page-about-media-right.php
carolinan:
```suggestion
					<h2 style="font-size:clamp(3rem, 6vw, 4.5rem);font-weight:300;line-height:1.115"><em>' . wp_kses_post( __( 'Emery<br>Driscoll', 'twentytwentytwo' ) ) . '</em></h2>
```

carolinan:
> > Do we also need to replace ’ as in Oh hello. My name’s Angelo, with ’ ?
> 
> What would be the reason for doing this? The current way results in the correct right singular apostrophe.

I remember needing to do that for TT1, but perhaps it was only inside HTML attributes.

jffng (approved):

PR #258, opened 2021-11-19
Fix duplicate block pattern

kjellr:
Our "About page with big image and buttons" pattern is showing up as a duplicate in the footer category by mistake, with the title "Footer with title, tagline, and social links". I think this was an error from duplicating files back when I was creating new patterns:

Screen Shot 2021-11-19 at 10 27 17 AM

This PR removes that pattern, and tidies up the naming of the actual "Footer with title, tagline, and social links" pattern. It also renames "About page with big image and buttons" to use the word "large" instead of "big" because that's more consistent with language we're using elsewhere.

Finally, since I was editing it anyway, I moved the "About page with big image and buttons" pattern down the list so that it's next to our other "link in bio"-style patterns.


inc/patterns/footer-title-tagline-social.php
```diff
@@ -1,84 +1,26 @@
 <?php
```
kjellr, line 1: This diff looks deceptively complex. This pattern is being replaced with the one that was previously in the `footer-title-tagline-social-dark.php` file, so that's why it looks like so much has changed.

inc/block-patterns.php
```diff
@@ -88,12 +87,12 @@ function twentytwentytwo_register_block_patterns() {
 		'hidden-404',
 		'hidden-bird',
 		'hidden-heading-and-bird',
-		'page-about-big-image-and-buttons',
 		'page-about-media-left',
 		'page-about-simple-dark',
 		'page-about-media-right',
 		'page-about-links',
 		'page-about-links-dark',
+		'page-about-big-image-and-buttons',
```
jffng: Should the file be renamed to `page-about-large-image-and-buttons` for consistency?

jffng (approved):
Pending my one suggestion, this LGTM.

inc/block-patterns.php
kjellr: Good call! I'll make that change and merge. 👍

PR #260, opened 2021-11-22
Try new navigation block markup

### LINKED ISSUE #230: Update Navigation block in templates to use layout attributes

The Navigation block was updated in https://github.com/WordPress/gutenberg/pull/36169 to use the new layout controls, such as Query loop, Row, Buttons and Social Icons blocks also use. This means that the attribute structure in Navigation block markup has changed slightly, so it would be good to update the Navigation blocks in Twenty Twenty Two templates accordingly.

Currently we have e.g.: `<!-- wp:navigation {"itemsJustification":"right","isResponsive":true} -->` . With layout, this should change to `<!-- wp:navigation {"layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right"}} /-->`. (`is-responsive` is now on by default so it's no longer necessary to specify it.)

Nothing should break if using the old attributes, because there is a deprecation in place to handle them, but as this is the default theme shipping with 5.9 it would be nice to use the latest markup 🙂

kjellr:
Closes #230.

Swaps out our existing navigation block markup to use layout attributes, and to remove `isResponsive` (that is now enabled by default, so there's no need for us to declare it). 

While doing this, I realized that the `header-centered-logo` pattern did not have a page list block inside of it like all of the others. So I added that in too. 

To test, just make sure all patterns appear like they did before, and also show no markup errors.

### DIFF
--- a/inc/patterns/page-about-simple-dark.php
+++ b/inc/patterns/page-about-simple-dark.php
@@ -6,9 +6,9 @@ return array(
 	'title'      => __( 'Simple dark about page', 'twentytwentytwo' ),
 	'categories' => array( 'twentytwentytwo-pages' ),
 	'content'    => '<!-- wp:cover {"overlayColor":"foreground","minHeight":100,"minHeightUnit":"vh","contentPosition":"center center","align":"full","style":{"spacing":{"padding":{"top":"max(1.25rem, 8vw)","right":"max(1.25rem, 8vw)","bottom":"max(1.25rem, 8vw)","left":"max(1.25rem, 8vw)"}}}} -->
-					<div class="wp-block-cover alignfull has-foreground-background-color has-background-dim" style="padding-top:max(1.25rem, 8vw);padding-right:max(1.25rem, 8vw);padding-bottom:max(1.25rem, 8vw);padding-left:max(1.25rem, 8vw);min-height:100vh"><div class="wp-block-cover__inner-container"><!-- wp:navigation {"layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right"},"overlayMenu":"always"} -->
-
-					<!-- wp:navigation {"itemsJustification":"right","overlayMenu":"always"} -->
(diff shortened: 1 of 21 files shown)


inc/patterns/page-about-simple-dark.php
```diff
@@ -6,7 +6,9 @@
...
 	'categories' => array( 'twentytwentytwo-pages' ),
 	'content'    => '<!-- wp:cover {"overlayColor":"foreground","minHeight":100,"minHeightUnit":"vh","contentPosition":"center center","align":"full","style":{"spacing":{"padding":{"top":"max(1.25rem, 8vw)","right":"max(1.25rem, 8vw)","bottom":"max(1.25rem, 8vw)","left":"max(1.25rem, 8vw)"}}}} -->
-					<div class="wp-block-cover alignfull has-foreground-background-color has-background-dim" style="padding-top:max(1.25rem, 8vw);padding-right:max(1.25rem, 8vw);padding-bottom:max(1.25rem, 8vw);padding-left:max(1.25rem, 8vw);min-height:100vh"><div class="wp-block-cover__inner-container"><!-- wp:navigation {"itemsJustification":"right","overlayMenu":"always"} -->
+					<div class="wp-block-cover alignfull has-foreground-background-color has-background-dim" style="padding-top:max(1.25rem, 8vw);padding-right:max(1.25rem, 8vw);padding-bottom:max(1.25rem, 8vw);padding-left:max(1.25rem, 8vw);min-height:100vh"><div class="wp-block-cover__inner-container"><!-- wp:navigation {"layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right"},"overlayMenu":"always"} -->
```
jffng, line 9: Getting a block validation error on this one, it looks like there are two `wp:navigation` blocks declared in the markup.

kjellr: Good catch — mind the-testing that one?

jffng (approved):

Issue #264, opened 2021-11-29
Update theme folders to "templates" and "parts"


kjellr:
Now that https://github.com/WordPress/gutenberg/pull/36647 has been merged, we should update the the following folder names: 

- `block-templates` ➡️  `templates`
- `block-template-parts`  ➡️  `parts`

kjellr:
@mburridge has volunteered to take this on. 🙌

ndiego:
I believe this can be closed.

kjellr:
Yep, all set. Thanks everyone!

PR #265, opened 2021-11-29
Changes to directory names

mburridge:
Changes the following directory names, as per issue #264 :

- `block-templates` ➡️ `templates`
- `block-template-parts` ➡️ `parts`


kjellr (approved):
Thanks! This works in my testing. 🙌

kjellr:
After discussing with @jffng, we're going to hold off on merging for just a moment since this feature isn't merged into Beta 1 yet. Should be ready in a day or so!

kafleg:
We are not going to change the name under Appearance?
See this screenshot: https://i.rankmath.com/ViRosf

tinjure20:
I was also wondering how about use in templates markup (example: [here](https://github.com/WordPress/twentytwentytwo/blob/trunk/block-templates/index.html#L1))?

carolinan:
> I was also wondering how about use in templates markup (example: [here](https://github.com/WordPress/twentytwentytwo/blob/trunk/block-templates/index.html#L1))?

The markup uses the source and slug for the block, wp:template-part, and the name of the template part block has not been changed, so we don't need to change anything there.

carolinan:
> We are not going to change the name under Appearance? See this screenshot: https://i.rankmath.com/ViRosf

I don't think so, there were discussions a few months ago, but not recently.

kjellr:
The [release bundled here](https://github.com/WordPress/twentytwentytwo/releases/tag/v0.4.0) has been committed to core for Beta 1, so I think we're good to merge this. 

I think this'll mean folks need to test this GitHub edition of the theme against Gutenberg `trunk` until 12.1 comes out, but that should be fine since folks can still test the version bundled with Beta 1.  

Thanks, everyone!

PR #275, opened 2021-12-03
Try a subtle opacity change on button hovers

### LINKED ISSUE #99: Buttons need hover, active, and focus states

kjellr:
Fixes #99.

Tries a subtle opacity adjustment for button hovers. It's a single CSS rule, generally works ok, and will mostly still let folks choose hover colors when/if that feature gets rolled into core. From a pure visual design perspective, it's not my first choice but it looks acceptable. If we're implementing button hovers, I think this is our best bet. 

The only real downside is that it's not totally obvious that there's a hover state for outlined buttons on light backgrounds. But it still might be better than what we've got (which is no hover state at all). 

## Screenshots
buttons
buttons-2

### DIFF
--- a/style.css
+++ b/style.css
@@ -65,6 +65,14 @@ a:active {
 	padding: calc(.667em + 2px) calc(1.333em + 2px);
 }
 
-.wp-block-file a.wp-block-file__button:hover {
-	opacity: 1;
-}
+/*
+ * Button hover styles.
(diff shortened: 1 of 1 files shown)


jffng (approved):
Design-wise, this feels better than nothing.

Also it's a small amount of fairly harmless CSS. So I think it's an improvement overall.

PR #276, opened 2021-12-06
Update birds video

kjellr:
Related to #150.

This PR adjusts the `birds.mp4` video to be just one second long and re-compresses it. 

It's a placeholder video anyway, so I think this fine. It brings the file size down to about a quarter of its original size (from `1.9mb` to `469k`). Here's how it looks now: 

https://user-images.githubusercontent.com/1202812/144878650-789a879d-a3fc-4050-9c03-135990356a46.mp4


scruffian (approved):
LGTM

jffng (approved):
Works, nice optimization.

PR #277, opened 2021-12-06
Use get_theme_file_path to load block patterns

### LINKED ISSUE #221: Stop using `__DIR__` to load block pattern files
When reviewing the theme for merge into Core, I noticed that [block patterns are being included](https://github.com/WordPress/twentytwentytwo/blob/99db6063beb87eda7f58c3983992d94cd756d9c2/inc/block-patterns.php#L101-L106) using `require __DIR__ . '/patterns/' . $block_pattern . '.php'`.

I was curious if there was a reason for this. If this gets switched to using something like `get_stylesheet_directory()`, or `get_template_directory()` (these may not do exactly what we need), it could become possible for a child theme to include one pattern override file in `patterns` directory.

jffng:
**Description**

This PR solves #221. It allows a child theme to override a pattern by including a file at the same location. This way, updates to the other patterns would still be preserved. 

**Testing Instructions** 

Ensure all the patterns still load as expected in TT2.

Bonus:

1. Create a twentytwentytwo child theme and activate it
2. Add a pattern with the same name (e.g. `inc/pattern/footer-default.php`)
3. Change something in the pattern, go to the post editor, ensure the editor loads the child theme's version of the pattern.

### DIFF
--- a/inc/block-patterns.php
+++ b/inc/block-patterns.php
@@ -124,7 +124,7 @@ function twentytwentytwo_register_block_patterns() {
 
 		register_block_pattern(
 			'twentytwentytwo/' . $block_pattern,
-			require $pattern_file 
+			require $pattern_file
 		);
 	}
 }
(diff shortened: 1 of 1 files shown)


aristath (approved):

kjellr (approved):
This works well in my testing. Thanks!

PR #279, opened 2021-12-07
Try adding a hover state for navigation items

kjellr:
This adds an underline hover state for nav items. It's just a few lines of code, and I think it's reasonable alongside the other hover state code we've implemented already. 

I've tested this with other nav block content that might contain links too (social links, site logo with link, etc). But the CSS specificity used here seems to work as expected.

Before|After
---|---
hover-default|hover-underline

### DIFF
--- a/style.css
+++ b/style.css
@@ -48,7 +48,8 @@ a:active {
 	text-decoration: none;
 }
 
-.wp-block-navigation .wp-block-navigation-item a:hover {
+.wp-block-navigation .wp-block-navigation-item a:hover,
+.wp-block-navigation .wp-block-navigation-item a:focus {
 	text-decoration: underline;
 	text-decoration-style: solid;
 }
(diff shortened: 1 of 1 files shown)


style.css
```diff
@@ -48,6 +48,11 @@ a:active {
 	text-decoration: none;
 }
 
+.wp-block-navigation .wp-block-navigation-item a:hover {
```
jffng: Are you intentionally leaving out the focus state here, since it could look a little strange?

For example, when the sub-menu item is in focus:
Screen Shot 2021-12-07 at 3 53 04 PM

jffng (approved):
I think this is fine to add, it doesn't look like this would interfere with any near term plans for the styling navigation block.

style.css
kjellr: Ah yes, we should add this to focus too. I think it looks fine: 

focus

PR #283, opened 2021-12-08
Change paths in patterns to use the template not the stylesheet

scruffian:
This is needed so that these images still work in child themes.


kjellr (approved):
This works for me, thanks @scruffian. It's basically an expanded version of #231 that helps consolidate the functions we're using for these onto `get_template_directory_uri()`.

PR #284, opened 2021-12-08
Try using the default categories

### LINKED ISSUE #267: Try and place patterns in default categories instead of declaring our own
Currently, Twenty Twenty-Two declares five new pattern categories: 

- Twenty Twenty-Two General
- Twenty Twenty-Two Footers
- Twenty Twenty-Two Headers
- Twenty Twenty-Two Posts
- Twenty Twenty-Two Pages

Most of these categories should/will/do exist in Gutenberg by default, and it would be great if we could hook into them automatically rather than declare our own duplicate, lengthy category names. 

For example, it would be ideal if we could just place our theme header patterns in the normal "Headers" category, and have them show up before the default set. 

## Screenshot

Screen Shot 2021-11-30 at 12 57 38 PM

kjellr:
Resolves #267.

This PR moves our patterns over to the default categories whenever it makes sense to do so. We reuse the following categories that already exist in core: 

- Featured
- Headers
- Query
- Buttons
- Columns
- Gallery
- Text

... and we create just a couple additional ones: 

- Footers
- Pages

When our patterns appear in pre-existing categories, they appear at the top of the list. 

Some patterns appear in multiple categories, which I think is fine. 

Some of our patterns have `general` in their filename, even though that category isn't used anymore. I think that's ok though, it's essentially just like calling them "miscellaneous". Open to objections though. 🙂

## Screenshots

Before: 
Screen Shot 2021-12-09 at 1 01 05 PM

After:

https://user-images.githubusercontent.com/1202812/145451797-bcd4e669-a580-4895-9bb7-c092959de024.mp4

### DIFF
--- a/inc/block-patterns.php
+++ b/inc/block-patterns.php
@@ -12,7 +12,7 @@
  */
 function twentytwentytwo_register_block_patterns() {
 	$block_pattern_categories = array(
-		'general' => array( 'label' => __( 'General', 'twentytwentytwo' ) ),
+		'featured' => array( 'label' => __( 'Featured', 'twentytwentytwo' ) ),
 		'footer' => array( 'label' => __( 'Footers', 'twentytwentytwo' ) ),
 		'header' => array( 'label' => __( 'Headers', 'twentytwentytwo' ) ),
 		'query'   => array( 'label' => __( 'Query', 'twentytwentytwo' ) ),
--- a/inc/block-patterns.php
+++ b/inc/block-patterns.php
@@ -12,11 +12,11 @@
  */
 function twentytwentytwo_register_block_patterns() {
 	$block_pattern_categories = array(
-		'twentytwentytwo-general' => array( 'label' => __( 'Twenty Twenty-Two General', 'twentytwentytwo' ) ),
-		'twentytwentytwo-footers' => array( 'label' => __( 'Twenty Twenty-Two Footers', 'twentytwentytwo' ) ),
-		'twentytwentytwo-headers' => array( 'label' => __( 'Twenty Twenty-Two Headers', 'twentytwentytwo' ) ),
-		'twentytwentytwo-query'   => array( 'label' => __( 'Twenty Twenty-Two Posts', 'twentytwentytwo' ) ),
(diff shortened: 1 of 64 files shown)


inc/block-patterns.php
```diff
@@ -12,11 +12,11 @@
...
 	$block_pattern_categories = array(
-		'twentytwentytwo-general' => array( 'label' => __( 'Twenty Twenty-Two General', 'twentytwentytwo' ) ),
-		'twentytwentytwo-footers' => array( 'label' => __( 'Twenty Twenty-Two Footers', 'twentytwentytwo' ) ),
-		'twentytwentytwo-headers' => array( 'label' => __( 'Twenty Twenty-Two Headers', 'twentytwentytwo' ) ),
-		'twentytwentytwo-query'   => array( 'label' => __( 'Twenty Twenty-Two Posts', 'twentytwentytwo' ) ),
-		'twentytwentytwo-pages'   => array( 'label' => __( 'Twenty Twenty-Two Pages', 'twentytwentytwo' ) ),
+		'featured' => array( 'label' => __( 'Featured', 'twentytwentytwo' ) ),
+		'footer' => array( 'label' => __( 'Footers', 'twentytwentytwo' ) ),
+		'header' => array( 'label' => __( 'Headers', 'twentytwentytwo' ) ),
+		'query'   => array( 'label' => __( 'Query', 'twentytwentytwo' ) ),
```
kjellr, line 18: I'm not a big fan of the "Query" name here, but it's used in Core right now so I think we should stick to it.

jffng (commented):
This is looking good when I test it against the 5.9 Beta.

However on Gutenberg trunk, I'm seeing an issue where the theme styles are not appearing in the preview. Maybe it's just me though?

Screen Shot 2021-12-09 at 1 33 12 PM

kjellr:
Hmm yeah, I see the same thing. Looking into it...

kjellr:
I'm seeing the same thing on the main branch of this repo, so I don't think it's a problem with this PR.

jffng (approved):

PR #290, opened 2021-12-10
Remove unnecessary font weight rule font-weight rule from index and home templates

kjellr:
This PR removes unnecessary font style + weight declarations from the `home.html` and `index.html` templates. These are unnecessary, because the same font weight rule is defined in theme.json already: 

https://github.com/WordPress/twentytwentytwo/blob/e9f49804eab30b51b9cd9b7d6b16674f80483b2c/theme.json#L232

... and `normal` is already the default value for this text anyway.

Removing these from the templates provides more flexibility in customization from child themes, and through `theme.json`. For example, [the "Swiss" variation in this PR](https://github.com/WordPress/theme-experiments/pull/292) tries to make all post titles bold. Before that rule was overridden here didn't work, but with this PR it works fine: 

Before|After
---|---
Screen Shot 2021-12-10 at 8 22 49 AM|Screen Shot 2021-12-10 at 8 25 08 AM

There should be no appearance change by default, since these deleted rules exist already.

### DIFF
--- a/inc/patterns/query-default.php
+++ b/inc/patterns/query-default.php
@@ -9,7 +9,7 @@ return array(
 	'content'    => '<!-- wp:query {"query":{"perPage":10,"pages":0,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":""},"align":"wide","layout":{"inherit":true}} -->
 					<div class="wp-block-query alignwide"><!-- wp:post-template {"align":"wide"} -->
 					<!-- wp:group {"layout":{"inherit":true}} -->
-					<div class="wp-block-group"><!-- wp:post-title {"isLink":true,"align":"wide","style":{"typography":{"fontStyle":"normal","fontWeight":"300"}},"fontSize":"huge"} /-->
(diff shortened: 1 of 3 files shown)


inc/patterns/query-default.php
```diff
@@ -9,7 +9,7 @@
 	'content'    => '<!-- wp:query {"query":{"perPage":10,"pages":0,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":""},"align":"wide","layout":{"inherit":true}} -->
 					<div class="wp-block-query alignwide"><!-- wp:post-template {"align":"wide"} -->
 					<!-- wp:group {"layout":{"inherit":true}} -->
-					<div class="wp-block-group"><!-- wp:post-title {"isLink":true,"align":"wide","style":{"typography":{"fontStyle":"normal","fontWeight":"300"}},"fontSize":"huge"} /-->
+					<div class="wp-block-group"><!-- wp:post-title {"isLink":true,"align":"wide","fontSize":"huge"} /-->
```
kjellr, line 12: For consistency, I made the same change in the pattern for this query layout.

jffng (approved):

PR #296, opened 2021-12-14
Fix alignment in footer blog pattern

kjellr:
The blog footer pattern is supposed to have its columns wide-aligned, but they're currently standard-width. This PR fixes that. 

Before|After
---|---
Screen Shot 2021-12-14 at 10 06 04 AM|Screen Shot 2021-12-14 at 10 06 37 AM


jffng (approved):
Fixes as described, thanks.

PR #298, opened 2021-12-14
Try updated font sizes

kjellr:
This PR changes our font sizes a bit, to align with the upcoming [font-size changes in Core](https://github.com/WordPress/gutenberg/pull/37381), and the work being done in https://github.com/WordPress/gutenberg/pull/37038. 

For Twenty Twenty-Two's purposes, we had to rename some of the size slugs: The "Normal" font size can't just be dropped outright since it's the base size used for the entire theme. So I changed that to Medium, and scaled things up from there. Here are the changes: 

Old preset sizes:

- Small
- Normal
- Medium
- Large
- Huge

New preset sizes:

- Small
- Medium
- Large
- Extra Large

Changes: 

- Small  ➡️ _(no change)_
- Normal ➡️ Medium
- Medium ➡️ Large
- Large ➡️ Extra Large
- Huge ➡️ _(custom variable, same name)_

### DIFF
--- a/inc/patterns/general-pricing-table.php
+++ b/inc/patterns/general-pricing-table.php
@@ -12,7 +12,7 @@ return array(
 					<!-- /wp:separator -->
 
 					<!-- wp:heading {"style":{"typography":{"fontSize":"var(--wp--custom--typography--font-size--gigantic, clamp(2.75rem, 6vw, 3.25rem))","lineHeight":"0.5"}}} -->
-					<h2 id="1" style="font-size:clamp(2.75rem, 6vw, 3.25rem);line-height:0.5">' . esc_html( _x( '1', 'First item in a numbered list.', 'twentytwentytwo' ) ) . '</h2>
+					<h2 id="1" style="font-size:var(--wp--custom--typography--font-size--gigantic, clamp(2.75rem, 6vw, 3.25rem));line-height:0.5">' . esc_html( _x( '1', 'First item in a numbered list.', 'twentytwentytwo' ) ) . '</h2>
 					<!-- /wp:heading -->
 
 					<!-- wp:heading {"level":3,"fontSize":"x-large"} -->
--- a/inc/patterns/general-pricing-table.php
+++ b/inc/patterns/general-pricing-table.php
@@ -11,7 +11,7 @@ return array(
 					<hr class="wp-block-separator is-style-wide"/>
 					<!-- /wp:separator -->
 
-					<!-- wp:heading {"style":{"typography":{"fontSize":"clamp(2.75rem, 6vw, 3.25rem)","lineHeight":"0.5"}}} -->
+					<!-- wp:heading {"style":{"typography":{"fontSize":"var(--wp--custom--typography--font-size--gigantic, clamp(2.75rem, 6vw, 3.25rem))","lineHeight":"0.5"}}} -->
 					<h2 id="1" style="font-size:clamp(2.75rem, 6vw, 3.25rem);line-height:0.5">' . esc_html( _x( '1', 'First item in a numbered list.', 'twentytwentytwo' ) ) . '</h2>
 					<!-- /wp:heading -->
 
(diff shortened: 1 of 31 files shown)


kjellr:
In bd39c49, I realized that we had to change the markup for all the items that used to be "Huge" too. Previously, they were using the `is-font-size-huge` class, which fell back to the Gutenberg default instead of our custom variable. 😕

I think I sorted that all out, but in any case, this could use a solid review if we move forward with it.

kjellr:
Ok, while I was here I made some very subtle refinements to the type scale so that it works better responsively.

This is a pretty large PR, but it's mostly find & replace, and I feel pretty confident about it. I'd like to merge this in sooner rather than later so that more folks are able to test it.

inc/patterns/general-pricing-table.php
```diff
@@ -11,12 +11,12 @@
 					<hr class="wp-block-separator is-style-wide"/>
 					<!-- /wp:separator -->
 
-					<!-- wp:heading {"style":{"typography":{"fontSize":"clamp(3rem, 6vw, 4rem)","lineHeight":"0.5"}}} -->
-					<h2 id="1" style="font-size:clamp(3rem, 6vw, 4rem);line-height:0.5">' . esc_html( _x( '1', 'First item in a numbered list.', 'twentytwentytwo' ) ) . '</h2>
+					<!-- wp:heading {"style":{"typography":{"fontSize":"clamp(2.75rem, 6vw, 3.25rem)","lineHeight":"0.5"}}} -->
```
jffng: Should these also use the custom variable with the `clamp` value as the fallback? e.g.

```suggestion
					<!-- wp:heading {"style":{"typography":{"fontSize":"var(--wp--custom--typography--font-size--gigantic, clamp(2.75rem, 6vw, 3.25rem))","lineHeight":"0.5"}}} -->
```

inc/patterns/hidden-heading-and-bird.php
```diff
@@ -10,8 +10,8 @@
 	'title'    => __( 'Heading and bird image', 'twentytwentytwo' ),
 	'inserter' => false,
 	'content'  => '<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"right":"max(1.25rem, 5vw)","left":"max(1.25rem, 5vw)","top":"0px","bottom":"0px"}}},"layout":{"inherit":true}} -->
-					<div class="wp-block-group alignfull" style="padding-top:0px;padding-right:max(1.25rem, 5vw);padding-bottom:0px;padding-left:max(1.25rem, 5vw)"><!-- wp:heading {"align":"wide","style":{"typography":{"fontSize":"clamp(4rem, 8vw, 6.25rem)","lineHeight":"1.15"}}} -->
-					<h2 class="alignwide" style="font-size:clamp(4rem, 8vw, 6.25rem);line-height:1.15">' . wp_kses_post( __( '<em>The Hatchery</em>: a blog about my adventures in bird watching', 'twentytwentytwo' ) ) . '</h2>
+					<div class="wp-block-group alignfull" style="padding-top:0px;padding-right:max(1.25rem, 5vw);padding-bottom:0px;padding-left:max(1.25rem, 5vw)"><!-- wp:heading {"align":"wide","style":{"typography":{"fontSize":"clamp(3.25rem, 8vw, 6.25rem)","lineHeight":"1.15"}}} -->
```
jffng: Same comment here about using the custom variable + fallback.

jffng (approved):
To confirm, I should I be seeing just four preset sizes in the typography panel, instead of the previous five: 

Screen Shot 2021-12-17 at 8 46 40 AM

Left a couple minor comments, once those are addressed I think we can bring this in.

jffng:
I did notice a bug on `trunk` which this PR fixes, but it could be an issue for other themes — the normal font size in the editor is being overridden by Gutenberg's default normal font size: 

Screen Shot 2021-12-17 at 8 50 23 AM

inc/patterns/general-pricing-table.php
kjellr: Yep, they should. 👍

kjellr:
> To confirm, I should I be seeing just four preset sizes in the typography panel, instead of the previous five

Thats right. 👍 

> I did notice a bug on trunk which this PR fixes, but it could be an issue for other themes — the normal font size in the editor is being overridden by Gutenberg's default normal font size:

I think it was doing that before, right? In any case, I'll leave a note in the corresponding Gutenberg PR to double check.

PR #301, opened 2021-12-15
Add menu container padding

### LINKED ISSUE #292: Close button placement does not align with the placement of the menu button
By default, when you click the menu button on a smalls screen, the location of the close button is slightly different than the location of the menu itself: 

tt2-menu

This is a little annoying, and it would be great if we could fix it. However, it's complicated by the fact that we have no idea where the menu button is actually going to be located. If the site title were stacked, then the menu button will move down from here slightly. And for what it's worth, the user can place a menu button in the footer, or all the way over to the left, etc. So there's no way we'd be able to match it exactly every time, and we probably wouldn't even want to. 

But still, this state where it's _almost_ in the same place feels like a bug. 

I'm not sure there's a good solution to this, and if there were it would likely need to happen upstream. But I'm opening the issue here to start discussion. 

cc @jasmussen, since I bet you've thought about this already at some point.

kjellr:
Closes #292 (It's not perfect, but it's the best we can do at the moment)

This PR modifies the default padding for the menu container so that it is in sync with with the site's outer padding. 

This means that (if you use the default header, have a one-line title, and have no site logo), the close button will probably line up with the menu button. But even without that set of specific circumstances, I think this is a reasonable move so that the outer spacing matches the site. 

Before|After
---|---
space-old|space

### DIFF
--- a/style.css
+++ b/style.css
@@ -124,7 +124,9 @@ body > .is-root-container,
 
 /*
  * Responsive menu container padding.
- * This gives the responsive container the same padding as the site itself.
+ * This ensures the responsive container inherits the same
+ * spacing defined above. This behavior may be built into
+ * the Block Editor in the future.
  */
 
 .wp-block-navigation__responsive-container.is-menu-open {
--- a/style.css
+++ b/style.css
@@ -121,3 +121,16 @@ body > .is-root-container,
 	margin-right: auto !important;
 	width: inherit;
 }
+
+/*
+ * Responsive menu container padding.
+ * This gives the responsive container the same padding as the site itself.
+ */
+
+.wp-block-navigation__responsive-container.is-menu-open {
+	padding-top: var(--wp--custom--spacing--outer);
+	padding-bottom: var(--wp--custom--spacing--large);
+	padding-right: var(--wp--custom--spacing--outer);
+	padding-left: var(--wp--custom--spacing--outer);
+}
+


style.css
```diff
@@ -121,3 +121,16 @@ body > .is-root-container,
 	margin-right: auto !important;
 	width: inherit;
 }
+
+/*
+ * Responsive menu container padding.
+ * This gives the responsive container the same padding as the site itself.
```
kjellr: @jasmussen, do you think this will ever be something that happens by default? If so I can add a code comment here and point to an issue.

jasmussen: For a modal overlay that covers the screen, yes I think it makes sense for it to inherit a global padding variable, OR at least have that padding variable easily configured via theme.json (and either would be good for the purposes of block themes, right?)

I can't point to a ticket you could link here, though! I think there are some pieces to work out around how separate overlay menu editing will happen. My current mental model is that it can function a bit like how editing a template part in isolation does, and this feature will likely have to land before the theme.json integration.

By the way, https://github.com/WordPress/gutenberg/pull/37323 landed which should simplify some things for this PR (not that I think you need ot do anything different), and I applied the backport label for TT2's usage.

kjellr: Ok, thanks. I clarified the code comment a little bit, but I'm not sure if I should be more explicit: 

```
This ensures the responsive container inherits the same
spacing defined above. This behavior may be built into
the Block Editor in the future.
 ```

jasmussen: Seems good! It serves also partially as a reminder to our future selves. 👍 👍

kjellr:
This feels pretty good, and makes sense conceptually alongside the outer padding CSS already in the PR. Could use a review when someone has a moment.

jffng (commented):
In theory this change makes sense, but I'm encountering an issue where closing the modal reopens it, since the buttons are now in the same place:

https://user-images.githubusercontent.com/5375500/146581779-f7f07263-5b5c-4237-9363-76b40655c671.mov

Tested on Chrome and Safari iOS mobile. It's replicable in the browser emulator if you use the "mobile" touch event profile. Seems like something to track upstream but it makes me hesitant to merge this in the meantime.

jasmussen:
Ack. I have noticed the course button being a bit twitchy, almost as if it closes onmousedown rather than onmouseup. Something we need to fix at the source.

kjellr:
Ok, we can hold off on this one for now I think. We can push this back in after the holidays. 👍

chthonic-ds:
> encountering an issue where closing the modal reopens it, since the buttons are now in the same place:

Hopefully fixed by https://github.com/WordPress/gutenberg/pull/36837.

jasmussen:
Thank you @chthonic-ds, yes, the issue does appear to be fixed by WordPress/gutenberg#36837. I just tested again. You can't quite see when I click, but in this GIF I took great care to click and hold on the burger icon, then release. Same with the close button, where I intentionally positioned the cursor over the adminbar. In the past, clicking the close icon when over the adminbar I'd accidentally go to the admin profile page, because the item underneith also receives the click:
test

In other words, this PR should be good to go.

kjellr:
Yep, I can't reproduce either anymore. I'll merge this in for the next release. Thank you all!

PR #304, opened 2021-12-16
Update Query Loop patterns with less posts per page

ntsekouras:
Related: https://github.com/WordPress/gutenberg/issues/37072

Patterns are really hard to handle in terms of performance as they can contain many many blocks and we not only have to parse and render them, but also do some recursive checks about the allowed blocks settings.

While this is an issue we have to handle in general and in core, for now I think we need to 'lighten' the theme's `Query Loop` patterns regarding the number of posts it previews. The point of a pattern is to showcase some design and with some pre-selected blocks(the inner Post blocks, etc..). It's already opinionated about the number of posts, so we could just show the minimum number required to make the pattern show its point.

With the increasing number of patterns it's crucial to think of these performance problems when we create one - again this is something that core needs to find a way to handle better.

The problem is absolutely noticeable with this theme as the number of posts was really big, taking not only a really long time to preview them, but also (maybe)failing to preview them at all - I didn't wait too much time to see if they would eventually render. If you try to see `Query` patterns without this PR you can see it.

## Notes
I removed some extra props that were not needed like `displayLayout` when `list` was used.
I think we should do this for GB and core `Query` patterns as well.

### DIFF
--- a/inc/patterns/query-simple-blog.php
+++ b/inc/patterns/query-simple-blog.php
@@ -6,7 +6,7 @@ return array(
 	'title'      => __( 'Simple blog posts', 'twentytwentytwo' ),
 	'categories' => array( 'query' ),
 	'blockTypes' => array( 'core/query' ),
-	'content'    => '<!-- wp:query {"query":{"perPage":1,"pages":0,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":true},"layout":{"inherit":true}} -->
+	'content'    => '<!-- wp:query {"query":{"perPage":1,"pages":0,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":false},"layout":{"inherit":true}} -->
 					<div class="wp-block-query"><!-- wp:post-template -->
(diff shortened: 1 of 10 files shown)


inc/patterns/query-simple-blog.php
```diff
@@ -6,7 +6,7 @@
 	'title'      => __( 'Simple blog posts', 'twentytwentytwo' ),
 	'categories' => array( 'query' ),
 	'blockTypes' => array( 'core/query' ),
-	'content'    => '<!-- wp:query {"query":{"perPage":3,"pages":0,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":true,"perPage":10},"layout":{"inherit":true}} -->
+	'content'    => '<!-- wp:query {"query":{"perPage":1,"pages":0,"offset":0,"postType":"post","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":true},"layout":{"inherit":true}} -->
```
ntsekouras: Should we be consistent with `core/GB` and the rest of the patterns here about the `inherit` value of the `Query Loop` block?

kjellr: Yeah, I think we should. It's an oversight that it's not doing that currently.

ntsekouras: I updated it - it was one instance and also made some updates in the number for page patterns.

ntsekouras:
Noting that some concerns were raised [here](https://github.com/WordPress/gutenberg/pull/37475#issuecomment-996534835), so let's wait a bit for this one.

ntsekouras:
I'm closing this for now due to the above concerns.

Issue #270, opened 2021-12-02
Questions and feedback: Downloading and installing the theme.


paaljoachim:
I made a new local site with Desktop Server to test out WP 5.9.

I downloaded Twenty Twenty Two version 0.4.0 from: https://github.com/WordPress/twentytwentytwo/releases/ (Please add the link directly to the readme.)

I see this:

![Screenshot 2021-12-02 at 11 28 15](https://user-images.githubusercontent.com/5323259/144404662-b0f248d5-b0b7-4b83-96e4-874b8d04f71f.jpg)

I click "Live Preview" to check out the site in the Customizer. 

![Screenshot 2021-12-02 at 11 34 34](https://user-images.githubusercontent.com/5323259/144405597-6f4242f2-dd72-47ce-81ef-ae0d1f52ab44.jpg)

The above feels a bit strange. One is able to click the Live Preview link, make modifications, but one is not able to activate the theme. 

I clicked X to exit.
Went back to the Themes page. Activated Twenty Twenty Two.
Noticed this message:

![Screenshot 2021-12-02 at 11 40 14](https://user-images.githubusercontent.com/5323259/144406433-006b2273-3240-4f42-aeac-e55b8a719995.jpg)

Clicked the back arrow.

Went to install and activate the Gutenberg plugin. (12.0.1)

Went back to activate the theme.
Theme activated. I clicked Live Preview. The Customizer opened and I saw this:

![Screenshot 2021-12-02 at 11 43 12](https://user-images.githubusercontent.com/5323259/144406844-c6ee6dd1-5656-451a-b48a-5332e260e5cc.jpg)

The Appearance menu shows this:

![Screenshot 2021-12-02 at 11 44 55](https://user-images.githubusercontent.com/5323259/144407346-0698446c-291d-4add-98de-68a2f9813f29.jpg)

I deactivate the Gutenberg plugin.

View the frontend. See a blank screen. I open the Chrome tools to see if I notice any error, but nothing is showing up here.

![Screenshot 2021-12-02 at 11 51 34](https://user-images.githubusercontent.com/5323259/144408506-d2e7285a-0370-4171-a1b2-afcc9d5f8f9c.jpg)

Going to Appearance -> Editor screen I see this: 
(There are also errors in Chrome console)

![Screenshot 2021-12-02 at 12 01 07](https://user-images.githubusercontent.com/5323259/144410013-96a5dcb5-f9d7-40bf-a485-7f35df1d158a.jpg)

[...]

PR #306, opened 2021-12-16
Port back-compat code from twentytwentyone

schlessera:
As a follow-up to the discussion during the core dev chat on Dec.15, this PR ports over the backwards-compatibility code from `twentytwentyone`. This adds admin notices and tries to avoid loading the theme on incompatible WordPress versions.

### DIFF
--- a/inc/back-compat.php
+++ b/inc/back-compat.php
@@ -1,0 +1,90 @@
+<?php
+/**
+ * Twenty Twenty-Two: Backwards compatibility functionality
+ *
+ * Prevents the theme from running on WordPress versions prior to 5.8,
+ * since this theme is not meant to be backward compatible beyond that and
+ * relies on many newer functions and markup changes introduced in 5.8.
+ *
+ * @since Twenty Twenty-Two 1.0
+ */
+
+/**
+ * Display upgrade notice on theme switch.
+ *
+ * @since Twenty Twenty-Two 1.0
+ *
+ * @return void
+ */
+function twentytwentytwo_switch_theme() {
+	add_action( 'admin_notices', 'twentytwentytwo_upgrade_notice' );
+}
+add_action( 'after_switch_theme', 'twentytwentytwo_switch_theme' );
+
+/**
+ * Adds a message for unsuccessful theme switch.
+ *
+ * Prints an update nag after an unsuccessful attempt to switch to
+ * the theme on WordPress versions prior to 5.8.
+ *
+ * @since Twenty Twenty-Two 1.0
+ *
+ * @global string $wp_version WordPress version.
+ *
+ * @return void
+ */
+function twentytwentytwo_upgrade_notice() {
+	echo '<div class="error"><p>';
+	printf(
+	/* translators: %s: WordPress Version. */
+		esc_html__( 'This theme requires WordPress 5.8 or newer. You are running version %s. Please upgrade.', 'twentytwentyone' ),
+		esc_html( $GLOBALS['wp_version'] )
+	);
+	echo '</p></div>';
+}
+
+/**
+ * Prevents the Customizer from being loaded on WordPress versions prior to 5.8.
+ *
+ * @since Twenty Twenty-Two 1.0
+ *
+ * @global string $wp_version WordPress version.
+ *
+ * @return void
+ */
+function twentytwentytwo_customize() {
+	wp_die(
+		sprintf(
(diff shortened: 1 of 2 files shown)


kjellr:
👋 Hey thanks for opening the PR. Here's [a link to the slack discussion for reference](https://wordpress.slack.com/archives/C02RQBWTW/p1639601904289000) ([signup](https://make.wordpress.org/chat/) is required to view). 

I understand why this is necessary, but philosophically bundling this with the theme seems like a rather inconvenient approach — the issue of the theme failing on earlier versions is not specific to Twenty Twenty-Two. This compat file would be necessary for _all_ block themes. 

I understand that it's better than say... patching _all_ prior versions of WP (😅), but it still doesn't feel right to me. 

cc @mtias in case you have any perspective here too. 

(Also potentially related: #270)

inc/back-compat.php
```diff
@@ -0,0 +1,90 @@
...
+	printf(
+	/* translators: %s: WordPress Version. */
+		esc_html__( 'This theme requires WordPress 5.8 or newer. You are running version %s. Please upgrade.', 'twentytwentytwo' ),
+		esc_html( $GLOBALS['wp_version'] )
+	);
+	echo '</p></div>';
+}
+
+/**
+ * Prevents the Customizer from being loaded on WordPress versions prior to 5.8.
```
kjellr, line 47: @jffng and I plan on changing the required version to 5.9 for release. We've been keeping it at 5.8 for now while folks are still testing on that version. (In general, it won't run on 5.8 without the Gutenberg plugin, so we feel that 5.9 is a more honest minimum version).

schlessera: Yes, I had assumed it would only work on 5.9+ but then noticed that the README said 5.8+, so I sticked with that for now. Using 5.9 now would probably make it untestable on current trunk.

schlessera:
> I understand why this is necessary, but philosophically bundling this with the theme seems like a rather inconvenient approach — the issue of the theme failing on earlier versions is not specific to Twenty Twenty-Two. This compat file would be necessary for all block themes.

Yes, I fully agree. Unfortunately, the extension interfaces of WordPress are ... non-existent - it just executes PHP files in certain filesystem locations. So this approach is used all over the place in both themes and plugins, where it is not WP Core that keeps things working but rather everyone else. 🤦

inc/back-compat.php
```diff
@@ -0,0 +1,90 @@
+<?php
```
jffng, line 1: I assume this file need to be included / called in `functions.php` somewhere.

jffng (commented):
Thanks for the PR. Like @kjellr, I don't think this shouldn't be necessary for the theme to add, once we bump the required version to 5.9. 

That said, what's the use case this should be tested with? Having the theme installed and then downgrading to <5.8?

inc/back-compat.php
schlessera: Yes, of course it does. 🤦

aristath (approved):
LGTM 👍 (provided of course that `5.8` is changed to `5.9` as mentioned above).

jffng:
After some more discussion (https://wordpress.slack.com/archives/C02RP4VMP/p1641222331449800 ) and testing, I don't think we should add this to the theme. 

In addition to the points raised in the discussion above, I can't recreate the scenario this is supposed to protect against. Here's what I tried:

- Upgraded to 5.9RC3 using the beta tester plugin
- Activated TT2
- Downgraded to 5.8.2
- Visited the themes page where you can see this:
Screen Shot 2022-01-19 at 10 45 08 AM
- Visited the customizer where I see this:
Screen Shot 2022-01-19 at 10 48 25 AM

With or without this code, I get the same result and notices.

schlessera:
May I ask how you downgraded to version 5.8.2 in the above test? Not all downgrades will happen in such a way that the core updater will execute its logic or trigger any hooks.

jffng:
> May I ask how you downgraded to version 5.8.2 in the above test?

Downgraded via the Beta Tester plugin.

kjellr:
As per the notes in https://github.com/WordPress/twentytwentytwo/pull/306#issuecomment-1016608252, I'm going to close this issue for now. If anyone would like to continue discussion, feel free to open [a new issue over on Trac](https://core.trac.wordpress.org/newticket?component=Bundled+Theme). 

Thanks!

PR #307, opened 2021-12-17
Fix Copyright date in style.css and readme.txt

kafleg:
As this theme will be released only in 2022 so need to fix the date.


kjellr:
I’m not sure about this one. The theme was designed, announced, and (primarily) developed in 2021, so I think the copyright should begin in that year.

jffng:
My understanding of the copyright date is how far back a copyright claim can be made. Since the theme was designed + developed in 2021, I also think 2021 makes sense.

PR #308, opened 2021-12-17
Change the default date value of pattern

kafleg:
Date changed to sync the released date.


kjellr:
Thanks for the PR, @kafleg. Beta 3 (packaged on Tuesday of this week) was the soft string freeze, and the theme has already been announced as [ready for translation](https://make.wordpress.org/polyglots/2021/12/16/wordpress-5-9-ready-to-be-translated/). Since this isn't a must-have change, I think we should just keep the string as-is.

Issue #309, opened 2021-12-20
Use --wp--custom--spacing--small variable instead of hard coded style


richtabor:
**Is your feature request related to a problem? Please describe.**
If I modify the `settings.custom.spacing.small` value (from the initial `max(1.25rem, 5vw)` value — then content will not be displayed appropriately throughout the theme, as a number of spacing values are tied are set specifically to `max(1.25rem, 5vw)` — instead of the small custom spacing setting in theme.json.

## Examples:
Here's the theme — **with no change** to the `settings.custom.spacing.small` default value:

![CleanShot 2021-12-20 at 17 29 37@2x](https://user-images.githubusercontent.com/1813435/146841231-e520b74b-617e-4a5e-9784-5b03987cbfad.png)

And here's the same page, with `settings.custom.spacing.small` set to `max(1.5rem, 10vw)` (doubled). 

![CleanShot 2021-12-20 at 17 26 11@2x](https://user-images.githubusercontent.com/1813435/146840912-165d7764-5f2d-4ae5-bc0c-1fdef3eba751.png)

In the screenshot above, the `header-large-dark.php` pattern has left/right padding values set to `max(1.25rem, 5vw)` within the pattern (instead of using the `--wp-custom--spacing--small` variable) — [source](https://github.com/WordPress/twentytwentytwo/blob/trunk/inc/patterns/hidden-heading-and-bird.php#L12). 

## Describe the solution you'd like:

There's not really a clean way for me to make the change without diving into the pattern markup on the page, so in essence this hard-coding of the padding value within the markup is stopping variants from being able to modify the small/outer spacing values. 

If we used `var(--wp--custom--spacing--small, 1.25rem)` instead of `max(1.25rem, 5vw)` throughout the theme, then users could modify the spacing throughout the theme within theme.json, like this: 

<img width="658" alt="CleanShot 2021-12-20 at 17 38 54@2x" src="https://user-images.githubusercontent.com/1813435/146842070-b0265c24-4119-4fe0-bb1a-9a228de45306.png">

richtabor:
Realistically, we should evaluate where it makes sense to use var(--wp--custom--spacing--outer) instead of var(--wp--custom--spacing--small). 

As currently `--wp--custom--spacing--small`, ( or its value of `max(1.25rem, 5vw)`) is used everywhere.

richtabor:
I spun out #310 based on the findings here.

[...]

PR #310, opened 2021-12-20
Adjust padding to better support theme.json and outer padding

richtabor:
**Description**
Remove forced padding values on patterns, that are not allowing `settings.custom.spacing.outer` to function properly.

**Screenshots**
Header Large Dark, with increased outer spacing (with settings.custom.spacing.outer set to 4rem for illustrative purposes):
CleanShot 2021-12-20 at 18 02 24@2x

header-small-dark (with settings.custom.spacing.outer set to 4rem for illustrative purposes):
CleanShot 2021-12-20 at 18 11 18@2x

**Testing Instructions** 

_Provide steps for testing_

1. Modify the theme.json value for `settings.custom.spacing.outer` to anything else (4rem for example)
2. See how the contents within patterns appropriately adjust to the new value


richtabor:
Here are more headers (showing how a change in outer is reflected properly): 
CleanShot 2021-12-20 at 18 23 59@2x

richtabor:
These examples show what happens to the various patterns' padding when a value other than `var(--wp--custom--spacing--small, 1.25rem)` is set as settings.custom.spacing.outer. 

Here's the current PR — which does not actually change the value, but makes the patterns adapt to when there is a change: 

CleanShot 2021-12-20 at 19 27 05@2x
CleanShot 2021-12-20 at 19 27 34@2x

inc/patterns/footer-dark.php
```diff
@@ -6,8 +6,8 @@
 	'title'      => __( 'Dark footer with title and citation', 'twentytwentytwo' ),
 	'categories' => array( 'footer' ),
 	'blockTypes' => array( 'core/template-part/footer' ),
-	'content'    => '<!-- wp:group {"align":"full","style":{"elements":{"link":{"color":{"text":"var:preset|color|background"}}},"spacing":{"padding":{"top":"max(1.25rem, 5vw)","bottom":"max(1.25rem, 5vw)","left":"max(1.25rem, 5vw)","right":"max(1.25rem, 5vw)"}}},"backgroundColor":"foreground","textColor":"background","layout":{"inherit":true}} -->
-					<div class="wp-block-group alignfull has-background-color has-foreground-background-color has-text-color has-background has-link-color" style="padding-top:max(1.25rem, 5vw);padding-right:max(1.25rem, 5vw);padding-bottom:max(1.25rem, 5vw);padding-left:max(1.25rem, 5vw)"><!-- wp:group {"align":"wide","layout":{"type":"flex","justifyContent":"space-between"}} -->
+	'content'    => '<!-- wp:group {"align":"full","style":{"elements":{"link":{"color":{"text":"var:preset|color|background"}}},"spacing":{"padding":{"top":"max(1.25rem, 5vw)","bottom":"max(1.25rem, 5vw)"}}},"backgroundColor":"foreground","textColor":"background","layout":{"inherit":true}} -->
```
kjellr: From my reading of #309, it sounded like we should actually be using the `var(--wp--custom--spacing--small, 1.25rem)` everywhere, right? So it can pull from the variable? That generally makes sense, but I'm confused about why you're using a hardcoded value here instead?

richtabor: I wanted this PR to remove the left/right padding - which makes the outer padding work better with the existing patterns. 

I didn’t want to change too much at once - but yea those should be using variables as well.

kjellr (approved):
Thanks for doing the heavy lifting on this one, @richtabor. This works well in my testing. 

This is somewhat unrelated, but while testing this I noticed that the header and footer patterns didn't work correctly in the site editor because they don't have layout settings inherited. I added an extra wrapper to those while I was reviewing, and they now work work as expected. I'll merge this in before I start making the PR any larger. 😅 Thanks again!

PR #312, opened 2021-12-25
Fix: Quote block border issue with right align settings

### LINKED ISSUE #278: [Quote]: Border issues
**Describe the bug**
The theme includes the following settings for the Quote block in `theme.json`:

```
"core/quote": {
  "border": {
    "width": "0 0 0 1px"
  }
}
```

It works great in a default block layout (with left alignment). However, issues appear when changing alignment from left to right.

**To Reproduce**
1. Add Quote block to the post.
2. Keep default alignment.
3. Add another Quote block to the post.
4. Change text alignment to right in the 2nd Quote block. 
5. See issue.

**Expected behavior**
The block should have the same styling, except the location of a side border, when setting either left alignment or right alignment. 

**Screenshots**

Website (different border width):
website_preview

Editor (missing a side border):
editor_preview

delowardev:
**Description**

Currently, the border size of the quote block has been set only for the left side, so when someone changes the alignment to the right, it doesn't show the border on the right side. Instead of using `0 0 0 1px`, we can use `1px` only; Gutenberg will take care of showing border based on alignment setting.
 
**Screenshots**

https://user-images.githubusercontent.com/17360543/147378570-49c95717-6a84-4e27-b212-c05f778f925b.mov

**Testing Instructions** 

1. Insert Quote block
2. Change alignment

Fixes: #278

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -249,7 +249,7 @@
 			},
 			"core/quote": {
 				"border": {
-					"width": "0 0 0 1px"
+					"width": "1px"
 				}
 			},
 			"core/site-title": {


kjellr (approved):
Thanks, @delowardev. This doesn't completely solve our problem. It looks right in the editor: 

Screen Shot 2022-01-03 at 9 14 30 AM

... but the front-end doesn't show the correct width for the right-aligned quote: 

Screen Shot 2022-01-03 at 9 14 40 AM

That's a Gutenberg issue, and this is better than what we have today, so I think we should merge this anyway. I'll update https://github.com/WordPress/gutenberg/issues/37198 with this new information.

PR #316, opened 2022-01-03
Use --wp--custom--spacing--small instead of hard-coded spacing styles

### LINKED ISSUE #309: Use --wp--custom--spacing--small variable instead of hard coded style

### LINKED ISSUE #317: Consider using spacing.large value in place of hard-coded 8rem styles in patterns

This is related to #309 — which I've submitted a [PR](https://github.com/WordPress/twentytwentytwo/pull/316) for, to connect the small spacing value to the theme.json entry. In that PR, I'm using the spacing.small theme.json value, instead of hard-coded "small" spacing in patterns. 

We're also using `8rem` throughout the patterns — as a hard-coded value for large spacing.

 Let's consider adapting those values to be connected to theme.json?  We don't have a matching theme.json `settings.custom.spacing` entry for `8rem` exactly, but we do have: `"large": "clamp(4rem, 10vw, 8rem)"`

We should test if we can use `settings.custom.spacing.large` as `var(--wp--custom--spacing--large, 8rem)` in place of the hard-coded `8rem` values within the patterns, to achieve the same effect as that PR I submitted for the `small` spacing values. 

### Screenshot:
 CleanShot 2022-01-03 at 12 06 07@2x

richtabor:
**Description**

Closes #309 and closes #317 by using `var(--wp--custom--spacing--small, 1.25rem)` in place of the hard-coded `max(1.25rem, 5vw)` styles within patterns. This way if a user modifies the theme's theme.json `settings.custom.spacing.small` value, it is reflected throughout the theme.

**Testing Instructions** 

1. First - without this PR - add a pattern and change the  `settings.custom.spacing.small` value 
2. See no changes, as the `max(1.25rem, 5vw)` values are hard-coded into the patterns
3. With this PR active, add any of the changed patterns to a page
4. Change the `settings.custom.spacing.small` value in theme.json
5. See the spacing adapt on the page.


kjellr (approved):
This works well, thanks @richtabor!

PR #319, opened 2022-01-04
Use custom unit spacers in templates and patterns

### LINKED ISSUE #272: Empty gaps seen in the site editor.
**Describe the bug**

**To Reproduce**

**Expected behavior**

Between the paragraph blocks.

kjellr:
Fixes #272. 

⚠️ This PR is reliant on https://github.com/WordPress/gutenberg/pull/36186 being backported for the 5.9 release. 

When this PR is active, everything should look the same on the front-end and in the editor. Here's what's changed under the hood: 

- The space between posts is now uses custom units, and maps directly to the theme's `var(--wp--custom--spacing--large)` variable. Before, it was a static pixel number that approximated that value.
- The dark header templates and patterns now include spacers to adjust their bottom margin, instead of hardcoding a margin value. That means folks can edit that value via the UI now. 

## Screenshot
 
Screen Shot 2022-01-04 at 8 34 40 AM

### DIFF
--- a/inc/patterns/header-large-dark.php
+++ b/inc/patterns/header-large-dark.php
@@ -7,8 +7,8 @@ return array(
 	'categories' => array( 'header' ),
 	'blockTypes' => array( 'core/template-part/header' ),
 	'content'    => '<!-- wp:group {"align":"full","layout":{"inherit":true}} -->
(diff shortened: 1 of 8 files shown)


kjellr:
As per the notes in https://github.com/WordPress/gutenberg/pull/36186#issuecomment-1008538808, dynamic spacer units won't make it into the initial 5.9 release. We should likely port this over to Trac post-release so we can follow up when they become available in a minor release.

kjellr:
I'm going to close this PR for now, since it's out of date. But I'll we should follow up on this again for the next point release. 👍

PR #320, opened 2022-01-04
Add a search template

### LINKED ISSUE #256: Reconsider adding a search template
In the site editor, when adding a new template, the user has the option to select between search and front page:

Template selection in the site editor when clicking on "Add new" in the template list view

If they select search, they are met with a blank template.
I think it would be easier for users to edit an existing template, if users want to make changes to the search results page.

kjellr:
Fixes #256 

This PR adds a search template. It's exactly the same as our index template except it has a search field at the very top. 

Before, the search results page had no header at all, and there was no context that explained what was searched for. It felt like a bug. This new field shows the current query, so it provides some helpful context, and also allows them to search again. 

Before|After
---|---
Screen Shot 2022-01-04 at 9 06 26 AM|Screen Shot 2022-01-04 at 9 01 28 AM

### DIFF
--- a/templates/search.html
+++ b/templates/search.html
@@ -1,9 +1,7 @@
 <!-- wp:template-part {"slug":"header","tagName":"header"} /-->
 
 <!-- wp:group {"style":{"spacing":{"padding":{"bottom":"var(--wp--custom--spacing--large, 8rem)"}}},"layout":{"inherit":true}} -->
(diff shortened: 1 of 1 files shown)


danieldudzic (approved):
Works great!

kjellr:
Thanks @danieldudzic! After some second-guessing about the design, I've made the search field standard-width instead of full. But other than that this should be good to go. 👍 

Screen Shot 2022-01-04 at 9 20 59 AM

PR #322, opened 2022-01-05
Change Site Title in the footer to a paragraph instead of an H1

### LINKED ISSUE #321: H1 in the footer and H1 in the Header in addition to the H1 in the beginning of the page content
From an SEO point of view, this is not the best design if there are 3 H1s on the same page:

I suggest replacing H1 in the Header in the Footer to divs.

kjellr:
The change from #40 appears to have been lost at some point. This PR re-instates it. To test, confirm that the Site Title in the footer appears in a `p` tag instead of an `h1`.

Screen Shot 2022-01-05 at 12 12 40 PM

Closes #321


scruffian (approved):
LGTM

PR #325, opened 2022-01-10
Header templates: move group block margin to padding instead

### LINKED ISSUE #272: Empty gaps seen in the site editor.

kjellr:
Fixes #272. 

Since dynamic spacer block units will not be part of 5.9, this PR pursues an alternate fix for #272 that moves the unsupported margin values on the theme's dark headers to padding values on its (pre-existing) wrapper parent instead. 

This is a little hard to find (since it's a variable, it currently shows up as an empty field in the editor), but it does allow the values to be editable. So it seems like a slight improvement at least. 

## Screenshot: 

padding


kjellr:
Closing in favor of #327.

PR #326, opened 2022-01-10
Update required WordPress version to 5.9.

kjellr:
As per [the notes here](https://github.com/WordPress/twentytwentytwo/pull/306#discussion_r770605126), this PR updates the required WP version to 5.9 for the theme. 

I tested this alongside the current version of WordPress `trunk` (which is already marked as 5.9). It works fine and shows no errors. 

If you already have the the theme active on an earlier version of WordPress, it will continue to work with the Gutenberg plugin active, but you'll see this message: 

Screen Shot 2022-01-10 at 9 40 40 AM

If you deactivate the theme, you will not be able to re-activate it without upgrading to 5.9: 

Screen Shot 2022-01-10 at 9 41 54 AM

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -18,9 +18,7 @@ Whether you’re building a single-page website, a blog, a business website, or 
 == Changelog ==
 
 = 1.0 =
-* Released: December
-
-https://wordpress.org/support/article/twenty-twenty-two-changelog#Version_1.0
+* Released: January 25, 2022
 
 == Copyright ==
 
--- a/readme.txt
+++ b/readme.txt
@@ -1,6 +1,6 @@
 === Twenty Twenty-Two ===
 Contributors: wordpressdotorg
-Requires at least: 5.8
+Requires at least: 5.9
 Tested up to: 5.9
 Requires PHP: 5.6
 Stable tag: 1.0
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ To get started with development:
 ## Requirements
 
 - Gutenberg plugin (latest)
-- WordPress 5.8+
+- WordPress 5.9+
 - PHP 5.6+
 - License: [GPLv2](http://www.gnu.org/licenses/gpl-2.0.html) or later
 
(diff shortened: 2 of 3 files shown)


readme.txt
```diff
@@ -18,9 +18,7 @@ Whether you’re building a single-page website, a blog, a business website, or
 == Changelog ==
 
 = 1.0 =
-* Released: December
-
-https://wordpress.org/support/article/twenty-twenty-two-changelog#Version_1.0
+* Released: January 25, 2022
```
kjellr, line 21: I also updated this while I was in here. The release date was wrong, and that link doesn't exist.

readme.txt
```diff
@@ -1,6 +1,6 @@
 === Twenty Twenty-Two ===
 Contributors: wordpressdotorg
-Requires at least: 5.8
+Requires at least: 5.9
```
jffng, line 3: With this line, folks using a 5.9 Beta or RC version would get the warning that the theme is not compatible. 

If we change it to the following instead, it will allow folks using the beta testing plugin to still activate the theme, and prevent folks using anything lower than 5.9 to activate it.

```suggestion
Requires at least: 5.9-beta
```

Thoughts? 

We could remove the `-beta` before the theme is bundled if that's preferred also, or leave it. I don't see much harm in leaving it.

kjellr: Hmm interesting. I don't have a preference other than that it feels weird to note a beta version here. But if it works, then it works. I'll lean on your expertise here. If you think that's the best path forward for now, that's fine with me.

jffng:
After some [more consideration](https://wordpress.slack.com/archives/C02JUSF02TT/p1641834429042300), we'll wait until just before the 24 hour code freeze to update the required version string. This will allow for the widest range of folks to continue activating and testing the theme.

jffng:
Since the [dry run + 24code freeze is happening on Monday morning (11a EST)](https://make.wordpress.org/core/2022/01/20/wordpress-5-9-release-day-process/), I'm going to bring this in now.

jffng (approved):

PR #327, opened 2022-01-10
Try a static spacer underneath the header

### LINKED ISSUE #272: Empty gaps seen in the site editor.

kjellr:
Alternative to #325 and #319. 
Closes #272.

This PR adds a static spacer block under the header to replace the unsupported margin value that was there previously. Ideally, this should be a dynamic value, but this is the best we can do for the moment since that's not landing in 5.9. 

Using a static value produces more space than intended on mobile, but overall I think the usability enhancement makes it worthwhile for the initial release. This also makes for a more efficient transition to #325, once dynamic spacer units will be supported in a future minor release. I (somewhat regretfully) think this is the best path forward for the moment. 

---

**Editor:** 

Screen Shot 2022-01-10 at 11 36 08 AM

**Frontend:** 

(Front-end desktop views are identical to `trunk`.)

Before|After
---|---
Screen Shot 2022-01-10 at 11 33 34 AM|Screen Shot 2022-01-10 at 11 33 45 AM

Before|After
---|---
Screen Shot 2022-01-10 at 11 33 20 AM|Screen Shot 2022-01-10 at 11 29 13 AM

### DIFF
--- a/parts/header-large-dark.html
+++ b/parts/header-large-dark.html
@@ -1,6 +1,6 @@
 <!-- wp:group {"layout":{"inherit":true}} -->
-<div class="wp-block-group"><!-- wp:group {"align":"full","style":{"elements":{"link":{"color":{"text":"var:preset|color|background"}}},"spacing":{"padding":{"top":"0px","bottom":"0px"}}},"backgroundColor":"foreground","textColor":"background","layout":{"inherit":true}} -->
-<div class="wp-block-group alignfull has-background-color has-foreground-background-color has-text-color has-background has-link-color" style="padding-top:0px;padding-bottom:0px"><!-- wp:template-part {"slug":"header","tagName":"header","align":"wide"} /-->
+<div class="wp-block-group"><!-- wp:group {"align":"full","style":{"elements":{"link":{"color":{"text":"var:preset|color|background"}}},"spacing":{"padding":{"top":"0px","bottom":"var(--wp--custom--spacing--large, 8rem)"}}},"backgroundColor":"foreground","textColor":"background","layout":{"inherit":true}} -->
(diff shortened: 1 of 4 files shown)


parts/header-large-dark.html
```diff
@@ -1,7 +1,9 @@
...
-<div class="wp-block-group alignfull has-background-color has-foreground-background-color has-text-color has-background has-link-color" style="margin-bottom:var(--wp--custom--spacing--large, 8rem);padding-top:0px;padding-bottom:0px"><!-- wp:template-part {"slug":"header","tagName":"header","align":"wide"} /-->
+<div class="wp-block-group"><!-- wp:group {"align":"full","style":{"elements":{"link":{"color":{"text":"var:preset|color|background"}}},"spacing":{"padding":{"top":"0px","bottom":"var(--wp--custom--spacing--large, 8rem)"}}},"backgroundColor":"foreground","textColor":"background","layout":{"inherit":true}} -->
+<div class="wp-block-group alignfull has-background-color has-foreground-background-color has-text-color has-background has-link-color" style="padding-top:0px;padding-bottom:var(--wp--custom--spacing--large, 8rem)"><!-- wp:template-part {"slug":"header","tagName":"header","align":"wide"} /-->
```
kjellr, line 3: You'll notice that I'm adding some bottom padding here. The bird image is supposed to have some padding underneath it, but it was removed erroneously at some point. 🤷‍♂️

The padding [is still present in the pattern](https://github.com/WordPress/twentytwentytwo/blob/trunk/inc/patterns/header-large-dark.php#L12-L13), as it should be. So I'm just adding it back in to the template too. 

Before|After
---|---
Screen Shot 2022-01-10 at 11 33 20 AM|Screen Shot 2022-01-10 at 11 29 13 AM

jffng (approved):
I agree the spacer improves the usability, so I also prefer this approach to #325.

LGTM.

PR #328, opened 2022-01-10
Update copyright year to match core.

jffng:
**Description**

In https://core.trac.wordpress.org/changeset/52427, the copyright year for the theme was changed in Core. This PR syncs that change back to the development repo.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -24,7 +24,7 @@ https://wordpress.org/support/article/twenty-twenty-two-changelog#Version_1.0
 
 == Copyright ==
 
-Twenty Twenty-Two WordPress Theme, 2021 WordPress.org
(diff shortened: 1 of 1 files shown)


kjellr (approved):
This should probably be [updated in `style.css` too, right](https://github.com/WordPress/twentytwentytwo/blob/trunk/style.css#L16)? Other than that this is good to go.

jffng:
It looks like `style.css` for other default themes (see [twentytwentyone](https://core.trac.wordpress.org/browser/trunk/src/wp-content/themes/twentytwentyone/style.css#L18) only cites the copyright year in which it was developed 🤷 .

kjellr:
As far as I remember, none of the development cycles for the other default themes spanned more than one year. 😄

PR #339, opened 2022-01-19
Also allow stacking of negative margins in cover

erikjoling:
**Description**

A fully aligned block inside a fully aligned cover-block needs to have double negative margins, because it has double positive paddings. If this makes sense ^_^' 

Just like the group block.

**Testing Instructions** 

1. Add a fully aligned group-block inside a fully aligned cover block.
2. See double horizontal padding.

### DIFF
--- a/style.css
+++ b/style.css
@@ -126,8 +126,8 @@ body > .is-root-container > .wp-block-template-part > .wp-block-cover,
 .wp-site-blocks .wp-block-columns .wp-block-column .alignfull,
 .is-root-container .wp-block-columns .wp-block-column .wp-block[data-align="full"],
 /* We also want to avoid stacking negative margins. */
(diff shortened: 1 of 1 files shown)


kjellr:
Thanks, @erikjoling! This looks good: 

Before|After
---|---
Screen Shot 2022-01-20 at 12 02 50 PM|Screen Shot 2022-01-20 at 12 03 13 PM

@jffng and I aren't making any more bugfix commits in advance of the 5.9 code freeze on Monday, so we'll save this one and can push it in shortly after release. 

Thank you!

erikjoling:
No problem at all. Thank you, also for the great theme :)

jffng:
Migrated to a PR in core, thanks again @erikjoling ! https://github.com/WordPress/wordpress-develop/pull/2223

PR #347, opened 2022-01-24
Update readme with archive info.

jffng:
**Description**

Updates readme to point to trac, where any updates or issues should happen.


kjellr (approved):
All set, thanks!

PR #348, opened 2022-01-24
Try using __unstableLocation attribute to migrate over existing classic menu

kjellr:
This PR is a quick attempt to automatically migrate over a primary menu if one was set in a classic theme.

Technically, this seems to work in my testing, but it feels weird to lean on something that's clearly labeled "unstable." 😕

## To test: 

1. Start with a classic theme that uses a "Primary" menu location. (Most of the recent default themes do this). 
2. Assign a menu to that Primary location using the Customizer. 
3. Be sure to hit "Publish" to save your changes.
4. Switch to the Twenty Twenty-Two theme. 
5. Observe that the primary menu on the front end displays the same menu you chose in step 2.

## Video:

https://user-images.githubusercontent.com/1202812/150858000-74abb127-03df-4aa0-87c5-b0075bddab8e.mp4

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -27,6 +27,14 @@ if ( ! function_exists( 'twentytwentytwo_support' ) ) :
 		// Enqueue editor styles.
 		add_editor_style( 'style.css' );
 
+		// Add a menu location.
+		// Used only for migrating over classic menus.
+		register_nav_menus(
+			array(
+				'primary' => esc_html__( 'Primary menu', 'twentytwentytwo' ),
+			)
+		);
+
 	}
 
 endif;
--- a/parts/header.html
+++ b/parts/header.html
@@ -6,6 +6,8 @@
 <!-- wp:site-title {"style":{"typography":{"fontStyle":"italic","fontWeight":"400"}}} /--></div>
 <!-- /wp:group -->
 
-<!-- wp:navigation {"__unstableLocation":"primary","layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"right"}} /--></div>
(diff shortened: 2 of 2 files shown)


jffng (commented):
Thanks for trying this. In addition to the use of an "unstable" attribute, there are two other undesirable aspects of this I'm noticing:

1. The Appearance > Menus panel becomes accessible, meaning a user would have two ways of editing the navigation:

Screen Shot 2022-01-24 at 3 26 21 PM

2. The menu in the site editor is not displaying the same thing as the front, for me:

Site Editor | Front-end
--- | ---
Screen Shot 2022-01-24 at 3 24 20 PM | Screen Shot 2022-01-24 at 3 27 08 PM

functions.php
```diff
@@ -27,6 +27,14 @@ function twentytwentytwo_support() {
...
 		add_editor_style( 'style.css' );
 
+		// Add a menu location.
+		// Used only for migrating over classic menus.
+		register_nav_menus(
+			array(
+				'primary' => esc_html__( 'Primary menu', 'twentytwentytwo' ),
+			)
+		);
+
```
kjellr, line 37: In my testing, this only worked if I declared a nav menu here. 😕

jffng:
Noting that the first issue I mentioned may have been patched in 5.9, the above test site is using 5.8.2 + Gutenberg `trunk`.

kjellr:
Due to the problems noted above and the guidance in https://github.com/WordPress/gutenberg/issues/38166#issuecomment-1020730015, we'll close this for now. We may end up revisiting this approach later on once Gutenberg's tools better support it.

PR #349, opened 2022-01-26
Change Image size of ducks.jpg to 423 KB from 768 KB

kafleg:
You will see, **WARNING: assets/images/ducks.jpg is 768.2 KB in size. Large file sizes have a negative impact on website performance and loading time. Compress images before using them.**

This is not a big issue but the themes team recommends theme author put the images less than 500KB.


kjellr:
Sorry, I didn't see this before I submitted the revised screenshot over on Trac: https://core.trac.wordpress.org/ticket/54911

Anyway, that one is a little smaller (358kb), so let's use that one instead. I'll close this for now, but thank you for the PR!

